PR #18, opened 2022-04-19
do not hardcode bash path as bash is not always /bin/bash

rnagy:


jengelh:
I have changed it to /bin/sh now; that should exist everywhere.

PR #21, opened 2022-08-05
use shm_mkstemp(3) on OpenBSD where O_TMPFILE is not available

rnagy:
O_TMPFILE is not available on OpenBSD however we can use shm_mkstemp(3) to generate a temporary shared memory object that we can unlink right after creating it so that when all the file descriptors referencing it are closed, the object will be gone.

### DIFF
--- a/exch/emsmdb/ftstream_parser.cpp
+++ b/exch/emsmdb/ftstream_parser.cpp
@@ -22,6 +22,10 @@
 #include "ftstream_parser.h"
 #include "rop_processor.h"
 
+#if defined(__OpenBSD__)
+#include <sys/mman.h>
+#endif
+
 using namespace std::string_literals;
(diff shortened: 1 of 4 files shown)


jengelh:
The original code used a named file on disk.
The current code (since 84236c5c4d98726a4a34ae7c0d25b5d85b5f470c, a52e680b8b356b4b7c5a79e150382f7e0084feea) uses an anonymous file on disk.
Your change proposes to use a file in shared memory, which may not be such a good idea, as there is uncertainty about the size of the file, multiplied with however many users one is expecting to utilize the system at any one given time. What do you think?

jengelh:
9337ec9be0a09b023b0e90e5eb95e8f33546f31f & 2f2859b32e51c58cf8e90373bd25a179e37b01e4

PR #30, opened 2022-12-28
add missing braces around initialization of subobject

rnagy:
this shuts clang up with -Wmissing-braces


jengelh:
Consider this sample:
```c++
struct S { int x; };
struct T { S x; };
int main()
{
        T b = {1};
        T c{1};
}
```
This compiles without warnings on clang++, but the moment you change T to `struct T { S x, y; };`, it starts complaining. There is something about the list-init rules that is not consistent between g++ and clang++, and I can't tell which compiler is right.

PR #34, opened 2022-12-31
use pkg-config to get the ldap library requirements and fallback to header searching

rnagy:
OpenLDAP ships with a pkg-config file and on *BSD systems the files of openldap
is installed outside of default search paths, so relying on the pkg-config file is the
best way to go. This commit makes pkg-config the preferred way, while falling back
to the original header and library search because not everyone is shipping the
ldap.pc file.

This also adds an error message if the ldap library cannot be found.


jengelh:
That should be `ldap` not `lldap`.

PR #47, opened 2023-02-15
Add vmime library to installation documentation (doc/install.rst)

clique2015:


jengelh:
It's already present:
```
* libHX >= 4.3
* libiconv (OpenBSD only)
* libvmime >= 0.9.2.175 👈
* libzstd
* MariaDB Connector/C or compatible
```

clique2015:
The gromox github build file   (.github/workflows/build.yml)
 
installed libvmime here
https://github.com/grommunio/gromox/blob/f16d6a0c58c1a7d529c88e07c0abe792072f7db4/.github/workflows/build.yml#L17

and also vmime (kisli/vmime) here
https://github.com/grommunio/gromox/blob/f16d6a0c58c1a7d529c88e07c0abe792072f7db4/.github/workflows/build.yml#L34

during our compilation without installing both libraries, 
we got an error here
https://github.com/grommunio/gromox/blob/f16d6a0c58c1a7d529c88e07c0abe792072f7db4/lib/email/send.cpp#L61

jengelh:
vmime and libvmime refer to the same project (vmime.org / github.com/kisli/vmime).

[Debian and some 23 other distros](https://repology.org/project/libvmime/versions) decided to call their source package `libvmime`; [only 1 uses](https://repology.org/project/vmime/versions) `vmime`.
This was/is correct, because the vmime project provided tar archives by the name of `libvmime-0.6.3.tar.bz2` (2005) .. `libvmime-0.9.1.tar.bz2` on their website, and the archive contains a directory named `libvmime-0.9.1`. v0.9.2 is different because of the on-the-fly archive generation of GitHub, which leaves no option for a custom tarball name.

The workflow file installs libvmime-dev merely to pull in the transitive dependencies of it (libgsasl-dev/libgnutls-dev/libicu-dev), so that a rebuild of (lib)vmime gets approximately the same features without extra cmake parameters.

clique2015:
Very well understood now. Thanks

PR #68, opened 2023-09-23
Update kopano2grommunio.sh

deajan:
On kopano 11.0.2, I had `/var/lib/kopano/attachments` folder containing 256 folders from `00` to `ff`, hence the migration script wouldn't work.
Updated the attachment dir test for this case.

### DIFF
--- a/tools/kopano2grommunio.sh
+++ b/tools/kopano2grommunio.sh
@@ -402,7 +402,9 @@ if [[ MountKopanoAttachments -eq 1 ]]; then
     fi
 
     # test for Kopano attachment store, 10 directories 0..9 must exist, we look for 3 directories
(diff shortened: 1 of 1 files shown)


jengelh:
Check eliminiated altogether. If one never used attachments, no directories will exist in files_v2. gromox-2.13-2-g6914a57e3

deajan:
As far as I understand this check was done to make sure sshfs has properly mounted the attachment dir before trying to migrate from kopano, and mitigate the following issue:

If someone configures the wrong directory to mount in script (something like `/var/lib/kopano` instead of `/var/lib/kopano/attachments`), the script won't complain about the bogus directory.

So, without the check, will migrating a mailbox fail or just be executed "without attachments" ?

jengelh:
Without attachment and with plenty of warnings as it tries to open files that don't exist. You can (and should) use a dry run consisting of just kdb2mt (without piping to mt2exm) to see if there are excessive problems with a source mailbox.

deajan:
Thanks for the clarification ;)

PR #98, opened 2024-06-05
define process names for syslog

midasi:
When using syslog, the process name is currently empty. To fix this, the syslog ident has to be set explicitly.


jengelh:
Which libc and syslog implementations do you have running?

midasi:
We use musl libc on Alpine Linux

jengelh:
but with systemd-journald or..?

midasi:
No, with traditional syslog files. Alpine doesn't use systemd or journald

PR #125, opened 2025-01-07
github-action: replace '-j$(cat ...)' with MAKEFLAG

crpb:
the job 'cpucores' is just to see if there is a difference but i guess it won't 🙈 
> https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories

then i tried to run it with 24.04(because of those notices/warnings) but either i'm having no luck or did something wrong. anyhow, it looks like libHX has a problem with `runs-on: ubuntu-24.04`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,9 +9,11 @@ on:
 
 jobs:
   build:
-    runs-on:
-      - ubuntu-24.04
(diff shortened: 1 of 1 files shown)


jengelh:
gromox-2.38-55-gc87b41fe1

PR #127, opened 2025-01-12
gromox-cleaner.service

crpb:
gromox-cleaner: fixes
* disable PrivateNetwork
* split concatenation to two ExecStart='s
* As .here isn't working w/o distributed setup disable it by default
  The option you will have to set in the config will be
  /etc/gromox/gromox.cfg:cleaner_here=.here
  which isn't pretty.
  .here should work on Servers w/ an empty result in
  `grommunio-admin config server` aswell as we are ALWAYS ".here".

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,10 +1,15 @@
-Snapshot 2.38.37
+Snapshot 2.38.76
 ================
 
 Fixes:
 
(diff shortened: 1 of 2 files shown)


jengelh:
gromox-2.38-101-g483d10a9d
gromox-2.38-103-gd7a866f34

PR #142, opened 2025-03-31
Master

komarovsergey98:


jengelh:
I cherry-picked most of these. The source branch was updated with the leftovers I _didn't_ take. Will solve them differently.

Issue #160, opened 2025-08-26
Suppress out-of-office replies for messages delivered to groups


p-try:
Currently, when an email is sent to a distribution group (e.g., sales@company.de), recipients receive out-of-office (OOF) auto-replies from each individual group member who is away. This behavior can be confusing and noisy for senders, especially when multiple members are on vacation. 

As noted in the [Grommunio community forum ](https://community.grommunio.com/d/2233-suppress-out-of-office-replies-for-messages-sent-to-groups), this differs from the behavior of Microsoft Exchange, which suppresses OOF responses when a message is sent to a group address. 

This issue specifically concerns server-managed distribution groups (not client-side mailing lists). The existing OOF suppression logic does not currently account for messages addressed to such groups, leading to unwanted auto-replies. 

Expected Behavior:
Out-of-office replies should be suppressed when the original message is sent to a distribution group address. 

Proposed Solution:
Implement logic to detect if the recipient is a distribution group and suppress OOF auto-replies accordingly during message delivery. 

This enhancement improves user experience and aligns Gromox/Grommunio behavior more closely with established enterprise email standards.

deajan:
Did this ever get adressed ?
with gromox-3.8.9.m9df9502 I still see OOF messages when sending to a distribution group that contains a single user with an OOF message.

p-try:
It didn't, but my fork is up-to-date and includes this functionality:

https://github.com/p-try/gromox/tree/supported_suppress_oof_for_groups
based on gromox 3.8.9

I just closed my MR after it didn't seem to get much attention. I started to think I was the only one who really needed that feature, so I chose to quietly maintain my own fork.

deajan:
@p-try Your code looks lean to me at least.
@jengelh Would you mind giving your opinion on this ?

deajan:
@p-try Looks like your work has been merged in current prime branch ^^

PR #161, opened 2025-08-26
Suppress OOF auto-replies for messages sent to groups

p-try:
This PR introduces a change to suppress out-of-office auto-replies when an email is sent to a distribution group (e.g., sales@company.de).  #160


jengelh:
[internal note: see also GXL-613]

jengelh:
This modifies the existing mail and the suppress header carries forward to non groups as well, which it should not.

PR #223, opened 2026-02-17
Re-add missing property-tag assignments in commit a3ff02a (exmdb: abolish sprintf,strcpy calls)

umgfoin:
Fixes unreadable msg-content  (istore: E-2901: illegal proptype (6849h) serialized)


jengelh:
Merged as gromox-3.4-126-g 14c58356c

Issue #147, opened 2025-05-08
sqlite3 database checks


deajan:
Currently writing a gromox repair script at https://github.com/netinvent/gromox/blob/repair-script/tools/gromox-mbox-repair-tool.sh

While reading the gromox doc [here](https://docs.grommunio.com/kb/sqlite.html#recovery), I noticed that `gromox-http` should be disabled while doing a `pragma integrity_check;`.

Isn't the integrity check RO and should not affect gromox-http ?
Is it perhaps blocking the DB and gromox-http will fail somehow ?:

As side question, should the check / repair actions described in the doc also apply to `midb.sqlite3` and perhaps `tables.sqlite3` ?

Thanks for your insight.

jengelh:
Hm. The sentence was added after gromox-2.30 / parallel read support. One hypothesis coming to mind worth testing is that integrity_check copes (or not) with a concurrent writer which is actively modifying rows and whether it would find e.g. supposedly broken table indices.
I'd probably take a 100000-message mailbox (clone it), and run integrity_check at the same time as e.g. `mbop emptyfld -r /` and seeing if it barfs.

deajan:
Thanks, I'll try that.
What about midb.sqlite, is it worth doing the same checks/repairs ?

PR #246, opened 2026-05-20
Mailbox check / repair script

deajan:
Using this script since more than a year without any major trouble nor data loss.

Main functions:
- Iter over a single or all maildirs
- Check mailbox using gromox-mbck
- Repair mailbox if error detected
- Check exchange.sqlite3 via sqlite3
- Recover exchange.sqlite3 via sqlite3 and keep a backup copy of original file
- Check midb.sqlite3 via gromox-mkmidb --integrity
- Recreate midb.sqlite3 on error
- Takes care to stop / restart gromox services in case an sqlite3 file needs to be repaired

Perhaps it could be useful to add it to gromox.
Feel free to change headers / licence / whatever if needed.

Respects sqlite3 checks as per https://github.com/grommunio/gromox/issues/147


deajan:
Have been running this script on multiple grommunio systems to cope with incoherent data, comming from gromox itself or various power outages. Script works quite well, never lost any data, respects repair actions where grommunio needs to be shutdown.

All in all I think it would be nice for admins to have a single repair script instead of having to run multiple small tools, which is nice for specialized admins.
Would you consider merging it, or is there a better place for it somewhere out there ?

jengelh:
sh is awkward. The more one wants to do, the more awkward the syntax gets.

Overall, that script looks like it wants to be gromox-mbop. At that point, mbop should just be made to do it.

Choosing not to go with shell scripts.

deajan:
Fair enough, thanks for your viewpoint.

The most crucial missing part today would be gromox-mbck also being able to repair sqlite3 files (gromox-mbck repairs a mailbox, but sqlite3 still finds errors in said file).

The other part would be to allow iter over multiple mailboxes, hence the shell script.

I'll keep my script around until integrated tools handle those scenarios.

PR #249, opened 2026-05-22
EWS: improve Outlook for Mac compatibility

nourdineb-ops:
This PR improves Outlook for Mac compatibility with Gromox EWS.

Main changes:
- stabilize Outlook Mac folder hierarchy and recipient cache handling
- persist UpdateItem To/Cc/Bcc recipient changes on draft messages
- restore SendOnly / SendAndSaveCopy behavior
- ensure recipients are preserved before submission
- ensure sent messages are correctly moved to Sent Items
- add minimal GetAppMarketplaceUrl support required by Outlook Mac

Validation:
- Outlook for Mac can connect successfully
- draft creation/editing works
- message sending works
- recipients remain visible in Sent Items
- messages are correctly removed from Drafts after successful send
- external delivery validated with Gmail and Free
- tested on existing and newly created mailboxes

This PR was rebased onto current upstream/master and cleaned up before submission.


include/gromox/mapi_types.hpp
```diff
@@ -392,7 +392,10 @@ enum {
 	PRIVATE_FID_LOCAL_FAILURES = 0x1b,
 	PRIVATE_FID_SERVER_FAILURES = 0x1c,
 	PRIVATE_FID_CONVERSATION_ACTION_SETTINGS = 0x1d,
-	PRIVATE_FID_UNASSIGNED_START = 0x1e,
+	PRIVATE_FID_ARCHIVE = 0x1e,
```
jengelh, line 395: MAPI clients can and do generate folders they need on their own.
For example, if PR_ADDITIONAL_REN_ENTRYIDS were to be empty, we would not need FID_SYNC_FAILURES and Outlook would just create the folder on its own.

exch/ews/structures.cpp
```diff
@@ -1672,21 +1685,46 @@ sFolder tBaseFolderType::create(const sShape& shape)
...
+			folderType = SEARCH;
+		}else if (frClass) {
+   			 const char* dn = shape.get<char>(PR_DISPLAY_NAME);
+
+    			if (class_match_prefix(frClass, "IPF.Appointment") == 0)
+        			folderType = CALENDAR;
+    			else if (
+        			class_match_prefix(frClass, "IPF.Contact") == 0 ||
+        			(dn && (
+            				strcmp(dn, "Cache des destinataires") == 0 ||
```
jengelh: This does not look like it will work with a non-French client at all

grammmike:
internal ref: !374

nourdineb-ops:
Rebased onto current master and addressed both review comments:

**Non-French client issue** (`exch/ews/structures.cpp`): confirmed and fixed. The Recipient Cache detection only checked the English and French display-name translations, but `data/folder_names.txt` has 5 translated names for this folder (en/fr/de/es/it) - German, Spanish and Italian mailbox owners would never get their client-created folder recognized. Extended the check to all 5.

To be precise about when this display-name check actually matters: it's a fallback that only fires for mailboxes provisioned *before* `PRIVATE_FID_RECIPIENT_CACHE` existed, where Outlook creates its own copy of the folder client-side (generic `frClass`, not `IPF.Contact*`) and gromox has to recognize it without a reserved folder ID to key off of. Freshly provisioned mailboxes never hit this path at all - they go through the `PidTagFolderId == PRIVATE_FID_RECIPIENT_CACHE` check earlier in the same function, which is locale-independent by construction.

**Eager folder reservation/creation**: understand the concern about mirroring `PR_ADDITIONAL_REN_ENTRYIDS`-style lazy creation (Sync Issues/Conflicts/Local/Server Failures). I looked into this before replying: unlike those folders, `PR_ADDITIONAL_REN_ENTRYIDS` is a fixed 5-entry array (Conflicts/Sync Issues/Local Failures/Server Failures/Junk - see `exch/emsmdb/icsdownctx_object.cpp`) that this PR never touches, so Archive/Conversation History/Recipient Cache were never going to be advertised through it either way. The eager creation in `mkprivate.cpp` is what lets new mailboxes take the clean ID-based path above instead of ever needing the display-name fallback - removing it would make the fragile name-matching the *only* path for all mailboxes rather than a legacy-only fallback, which seems like a step in the wrong direction. Happy to reconsider if there'''s a cleaner mechanism you'''d prefer for the legacy-mailbox case specifically.

Force-pushed the rebased + fixed branch; PR now shows mergeable against current master.

jengelh:
About half the commits were cherry-picked, using the vcs.grommunio.com//gromox//347 internal resubmission. The parts that pertain to folder names and static folder ID reservations have been left out because unresolved in both #249 and vcs-347. Feel free to retry.

>MAPI clients can and do generate folders they need on their own.

PR #260, opened 2026-06-16
http: release parked HPM contexts on unexpected input

fmu83:
When an HPM context is parked, `http_parser::wait` currently only treats
`recv(MSG_PEEK) == 0` as a lost connection. In an Outlook/EWS StreamingEvents
scenario, stale parked connections were observed in `CLOSE-WAIT` while bytes
were still queued in `Recv-Q`. In that state `recv(MSG_PEEK)` returns a positive
value, so the parked context is not released and continues to occupy an HTTP
context slot.

After enough such contexts accumulate, gromox-http can reject or stall new EWS
requests until the service is restarted.

This change treats positive `MSG_PEEK` results while parked as unexpected input
and runs the context off. It also clears `hpm_wakeup_pending` when a context is
directly woken from `hsched_stat::wait`, so the wakeup latch is not consumed
again during the next park attempt.

Observed with Outlook Classic for Mac EWS sync:

- before the patch: growing `CLOSE-WAIT` count on gromox-http ports and Outlook
  stuck while determining changes;
- after the patch: repeated Outlook Mac/Windows sync cycles without new
  `CLOSE-WAIT` buildup on the gromox-http listener ports and without HTTP 503
  saturation.

### DIFF
--- a/exch/http/hpm_processor.cpp
+++ b/exch/http/hpm_processor.cpp
@@ -123,20 +123,17 @@ static void hpm_processor_wakeup_context(unsigned int context_id)
 {
 	auto phttp = static_cast<HTTP_CONTEXT *>(http_parser_get_contexts_list()[context_id]);
 	/*
-	 * Latch the wakeup *before* inspecting sched_stat. If the context is
-	 * mid-park (its plugin's retr() already returned HPM_RETRIEVE_WAIT, but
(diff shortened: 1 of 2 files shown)


fmu83:
issue with hanging macOS outlook classic is only partially fixed with this change.

PR #270, opened 2026-07-12
ews: fix FindPeople recipient autocomplete (Outlook)

nourdineb-ops:
## Summary

FindPeople is what Outlook (Mac, and other EWS clients) call live on every keystroke to autocomplete a recipient in the To/Cc/Bcc fields. On this deployment it never worked - every query returned no usable results, so the client always showed "person not found" even for addresses that definitely exist.

Two commits, found and fixed in that order (the first alone was not enough).

## `ews: fix FindPeople returning personas with no usable email/type`

- `FindPeople`/`GetPersona` read a persona'''s email via `fetch_prop(PR_SMTP_ADDRESS)`, which is never populated in ab_tree'''s generic propvals map - the match check silently never succeeded for anyone. Switched to the dedicated `user_info(ab_tree::userinfo::mail_address)` accessor, which `ResolveNames` already uses correctly for the same lookup.
- Added the missing `PersonaType` field - Outlook Mac silently discards a persona even with a valid email if this is absent.

## `ews: fix FindPeople response structure and add recipient-history search`

Even with the above fixed, the client still reported no results for every query. Root cause: `FindPeopleResponse` was wrapped in the `ResponseMessages`/`FindPeopleResponseMessage` batch envelope used by most other EWS operations, but FindPeople is not a batch operation - a response captured from a genuine Exchange server shows `ResponseClass` sits directly on the `FindPeopleResponse` root element, with `People`/`TotalNumberOfPeopleInView`/`FirstMatchingRowIndex`/`FirstLoadedRowIndex` as direct children, no wrapper. The client was silently discarding every response regardless of content correctness because of this extra nesting level.

Confirmed by MITM-injecting a real Exchange response body in place of gromox'''s own for one query - the client accepted it immediately, isolating the bug to response structure rather than content. Fixed by making `mFindPeopleResponse` inherit `mResponseMessageType` directly (dropping the separate `mFindPeopleResponseMessage` type), matching the flat pattern already used correctly by `mGetPersonaResponseMessage`/`mGetAppManifestsResponse` elsewhere in this codebase.

Also bundled into this commit:
- `PersonaId` (new `tPersonaId` type), `GivenName`/`Surname`, `RelevanceScore`, and a proper nested `EmailAddress` (Name/EmailAddress/RoutingType/MailboxType) on `tPersona` - the client silently drops personas missing these.
- A "Mailbox" query-source search over grommunio-web'''s `PR_EC_RECIPIENT_HISTORY_JSON`, so autocomplete also surfaces people actually emailed before, not just directory/GAL entries - reusing the same mechanism the webapp itself already relies on for its own (working) autocomplete.
- A real bug fix: `msg.People.emplace()` was called on every loop iteration in both the ab_tree match loop and the history-merge step; `optional<vector>::emplace()` destroys/reconstructs the vector each call, so only the last ab_tree match ever survived a response with multiple matches.

## Testing

Verified live against a real Outlook for Mac client (16.110.26070318): typing a partial name now correctly surfaces both directory and previously-emailed contacts in the autocomplete dropdown. Running in production since 2026-07-12.

### DIFF
--- a/exch/ews/requests.cpp
+++ b/exch/ews/requests.cpp
@@ -320,10 +320,27 @@ void process(mFindPeopleRequest &&request, XMLElement *response, const EWSContex
 				ab_tree::ab_node node(base.get(), mid);
 				tPersona persona;
 				std::string val;
+				/*
+				 * Outlook Mac silently discards personas with
+				 * no PersonaType, even when DisplayName/
+				 * EmailAddress are populated - it has no way
+				 * to classify the suggestion, so it never
+				 * offers it in the autocomplete list.
+				 */
+				persona.PersonaType = "Person";
 				if (node.fetch_prop(PR_DISPLAY_NAME, val) == ecSuccess)
 					persona.DisplayName = std::move(val);
-				if (node.fetch_prop(PR_SMTP_ADDRESS, val) == ecSuccess)
-					persona.EmailAddress = std::move(val);
+				/*
+				 * PR_SMTP_ADDRESS is never present in the
+				 * ab_tree node's generic propvals map (unlike
+				 * ResolveNames, which correctly uses this
+				 * dedicated accessor) - fetch_prop() here
+				 * always missed, leaving personas with no
+				 * usable address for Outlook to autocomplete.
+				 */
+				if (auto email = node.user_info(ab_tree::userinfo::mail_address);
+				    email != nullptr && *email != '\0')
+					persona.EmailAddress = email;
 				if (node.fetch_prop(PR_TITLE, val) == ecSuccess)
 					persona.Title = std::move(val);
 				if (node.fetch_prop(PR_NICKNAME, val) == ecSuccess)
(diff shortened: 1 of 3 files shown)


jengelh (commented):
>Also, on the way to finding that:

Quoting from an older version of the QEMU wiki:

"Also, ..." is often a good candidate for splitting into multiple patches.

If possible, please try this.

nourdineb-ops:
Good call, done - split the second commit into five:

- `ews: fix FindPeople only returning the last ab_tree match` (the pre-existing `optional<vector>::emplace()` bug)
- `ews: fix GetPersona returning personas with no usable email` (same PR_SMTP_ADDRESS issue as the first commit, but for GetPersona)
- `ews: add PersonaId/GivenName/Surname/RelevanceScore and structured EmailAddress to tPersona` (the new fields Outlook needs to accept a persona)
- `ews: un-batch-wrap FindPeopleResponse to match real Exchange shape` (the actual headline fix)
- `ews: search recipient history (grommunio-web mailbox JSON) for FindPeople matches` (the new "Mailbox" query source)

Each builds and links (`libgromox_ews.la`) on its own, and the resulting tree is identical to what was there before - just re-cut into smaller pieces.

jengelh:
It seems I already took these commits from, and by relying on the stacked nature of, #296 .

PR #271, opened 2026-07-12
mbop: add movemsg command

nourdineb-ops:
## Summary

Adds `movemsg`, a new gromox-mbop subcommand to move or copy a message between folders via exmdb'''s `movecopy_messages`, mirroring the existing single-purpose style of `delmsg`/`emptyfld`. Originally written as a manual test tool while validating an exmdb `message_moved` notification-subscription path, but it fills a real gap: mbop has no existing way to relocate a specific message between folders from the command line.

Second commit adds the man page section (matching the `delmsg`/`emptyfld` documentation style) and fixes the subcommand'''s own usage string, which advertised comma-separated message IDs (`message_id[,...]`) but the parser actually takes them as separate positional arguments, same as `delmsg`.

## Usage

```
gromox-mbop -u a@b.de movemsg -f src_folder_id -t dst_folder_id msgid...
gromox-mbop -u a@b.de movemsg -f src_folder_id -t dst_folder_id --copy msgid...
```

### DIFF
--- a/doc/gromox-mbop.8
+++ b/doc/gromox-mbop.8
@@ -62,6 +62,8 @@ get\-photo: retrieve user image from store and print to stdout
 get\-websettings, get\-websettings\-persistent, get\-websettings\-recipients:
 retrieve settings for grommunio-web
 .IP \(bu 4
+movemsg: move or copy a message between folders
+.IP \(bu 4
 ping: cause a mailbox's sqlite files to be opened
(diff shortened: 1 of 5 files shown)


jengelh:
added as gromox-3.8-102-g 5b8a77b20 .
If you can, you should set a proper author name (`git config user.name whoami`, `git config user.email here@there`).

PR #275, opened 2026-07-15
smtp: advertise SMTPUTF8 in delivery-queue's EHLO response

nourdineb-ops:
### Problem

gromox-delivery-queue (mda/smtp/*, the internal SMTP receiver that Postfix's `virtual_transport` relays local/virtual mail to, typically on `127.0.0.1:24`) never advertised `SMTPUTF8` in its EHLO capability list:

```
250-PIPELINING
250-HELP
250-SIZE 67108864
250 8BITMIME
```

Meanwhile Postfix's own public-facing smtpd/submission listeners advertise SMTPUTF8 by default. When an external sender negotiates SMTPUTF8 on the way in (or when Postfix otherwise tags a queued message as requiring it, e.g. via a mailbox forwarding rule re-injecting a message that already carries the flag), Postfix's outbound SMTP client refuses to relay that message onward to any next-hop that doesn't itself offer SMTPUTF8. Since gromox-delivery-queue never advertised it, every such message was **permanently bounced**:

```
dsn=5.6.7, status=bounced (SMTPUTF8 is required, but was not offered by host ::1[::1])
```

This reproduced reliably in production and caused real, user-facing mail loss (silent permanent bounce back to the original sender) for any message tagged SMTPUTF8, including plain internal-forward re-delivery.

Note `mda/lda.cpp` (a separate local-pipe delivery agent) already advertises `SMTPUTF8` correctly - this PR just brings the actual SMTP-listening delivery-queue daemon in line with it.

### Fix

Add `250-SMTPUTF8` to the EHLO response built in `mda/smtp/cmd.cpp`'s `cmdh_xhlo()`.

### Testing

Built and installed on a production host. Verified via a manual EHLO on the delivery-queue port before/after:

```
# before
250-HELP
250-SIZE 67108864
250 8BITMIME

# after
250-HELP
250-SIZE 67108864
250-8BITMIME
250 SMTPUTF8
```

End-to-end: sent a real message via `MAIL FROM:...  SMTPUTF8` to a local mailbox with an active forwarding rule (the exact path that was bouncing in production). Before the fix this bounced with `dsn=5.6.7`; after the fix, Postfix's log shows clean delivery:

```
to=<...>, orig_to=<...>, relay=::1[::1]:24, dsn=2.0.0, status=sent (250 Ok queue-id: 1)
```

### DIFF
--- a/mda/smtp/cmd.cpp
+++ b/mda/smtp/cmd.cpp
@@ -81,7 +81,8 @@ static int cmdh_xhlo(std::string_view cmd_line, smtp_context &ctx) try
 	string_length += gx_snprintf(&buff[string_length], std::size(buff) - string_length,
         "250-HELP\r\n"
 		"250-SIZE %zu\r\n"
-        "250 8BITMIME\r\n",
+        "250-8BITMIME\r\n"
+        "250 SMTPUTF8\r\n",
         /* send the size of "SIZE" command */
 		g_param.max_mail_length);
 


jengelh:
Advertising SMTPUTF8 also requires that we support UTF-8 addresses, which we don't.

```
mail from:<ht@a4.inai.de>
250 Ok
rcpt to:<用户@example.net>
550 unknown user - <=?utf-8?B?55So5oi3?=@example.net>
```

nourdineb-ops:
Fair point, and I can confirm it reproduces identically against our own build:

```
EHLO test.local
250 ... SMTPUTF8
MAIL FROM:<ht@a4.inai.de> SMTPUTF8
250 Ok
RCPT TO:<用户@example.net>
550 unknown user - <=?utf-8?B?55So5oi3?=@example.net>
```

Traced it: `EMAIL_ADDR::parse()` (`include/gromox/mail_func.hpp`) goes through vmime, which re-encodes a non-ASCII local part as an RFC 2047 encoded-word before it ever reaches the mailbox lookup - so a genuine UTF8SMTP mailbox name can never resolve correctly here. Agreed that's a real gap, and advertising SMTPUTF8 without it is misleading to a compliant sender.

For context on what this PR was actually fixing in production: the failing case we hit had *both* envelope addresses plain ASCII (an internal mailbox-rule forward, `a@ourdomain -> b@ourdomain`). Digging further just now, the SMTPUTF8 requirement on that message almost certainly wasn't about the addresses at all - `lib/ruleproc.cpp`'s `OP_FORWARD` resubmits via `sendmail://localhost` (vmime's sendmail transport, i.e. Postfix's local `sendmail` CLI), and Postfix's default `smtputf8_autodetect_classes = sendmail, verify` auto-tags any `sendmail`-submitted message as SMTPUTF8-required if it contains raw non-ASCII header content (e.g. an unencoded display name). So the actual bug is plausibly upstream of this: whatever `oxcmail_export` hands to the sendmail transport carrying non-ASCII header bytes without RFC 2047 encoding, tripping Postfix's autodetect for a message that never needed EAI/UTF8SMTP addressing at all.

I haven't confirmed that oxcmail_export angle with a byte-level capture yet, but if it holds up it'd be a cleaner fix - no address-EAI support required, and it wouldn't need this EHLO change at all. Given that, how would you like to proceed with this PR: hold it while I chase the oxcmail_export header-encoding angle instead, or is a properly-scoped SMTPUTF8 (with real UTF8 address support) something worth doing on top of this regardless? Happy to go either way, didn't want to guess.

nourdineb-ops:
You're right, and I'd rather withdraw this than paper over it.

I finally tracked down the actual root cause of the bounce that motivated this PR, and it has nothing to do with gromox. All of the original bounces (`SMTPUTF8 is required, but was not offered by host ::1[::1]`) turned out to come from a single internal host on my network running its own Postfix, submitting cron/monitoring mail via the local `sendmail` command. That host's `smtputf8_autodetect_classes` is at the Postfix default (`sendmail, verify`), which unconditionally tags any locally-submitted (`sendmail` CLI) message as SMTPUTF8-required regardless of whether the content is actually non-ASCII. That tag then rode along over a real SMTP hop into this server and only broke at the last leg, into `gromox-delivery-queue`, because that listener didn't advertise SMTPUTF8 - which is exactly correct given it doesn't support UTF8SMTP addressing, as you pointed out.

Confirmed by re-testing with the fix rolled back: genuine external mail (no SMTPUTF8 negotiated) delivers fine either way, and the only messages that ever tripped this were ones already carrying an SMTPUTF8 requirement they didn't actually need, from a misconfigured sender. The right fix is on that sender's end (drop `sendmail` from its `smtputf8_autodetect_classes`), not by having gromox claim UTF8SMTP support it doesn't have.

Closing this - thanks for pushing back on it rather than letting it merge.

PR #281, opened 2026-07-20
exmdb: fix inverted ST_ENABLED test for extended rules

r4nc0r:
b18cb1741 dropped the negation, so message_load_folder_ext_rules now skips every enabled extended rule and keeps only disabled ones -- the inverse of message_load_folder_rules, which the same series fixed correctly.


jengelh:
i fucking hate that section

PR #284, opened 2026-07-21
exmdb: fix new-message ordering in msgtime_index-sorted content tables

r4nc0r:
### Problem

In Outlook **online (non-cached) mode**, a mail arriving into an *open* Inbox sorted by Received date shows up at the bottom of the list and is briefly displayed twice; switching folders and back fixes it. Reported at https://community.grommunio.com/d/2381 (Outlook 2016/2021/2024, present since ~3.1, still in 3.4.80). Cached mode is unaffected because the client sorts locally.

### Root cause

A content table sorted by delivery/modification/submit time is loaded via the `msgtime_index` fast path added in `c104ecab5` (DESK-3893), which sets `table_node::psorts = nullptr`. `dbeng_notify_cttbl_add_row` treats a null `psorts` as "unsorted" and appends the new row at the bottom of the table. So a message delivered into an open time-sorted folder is placed at the end until the table is reloaded (which re-sorts correctly via `msgtime_index`). The mis-anchored `cttbl_row_added` notification vs. the client's own range queries is what produces the transient duplicate.

The uncategorised `stbl` comparison branch positions rows correctly but has no tiebreaker, so equal-key rows (large groups under non-time sorts) can also diverge from the reload order.

### Fix

- `table_node` remembers the accel sort key (`accel_tag`/`accel_dir`) even though `psorts` stays null for the fast load path.
- `dbeng_notify_cttbl_add_row`: the append-at-bottom branch now applies only to genuinely unsorted tables (`psorts == NULL && accel_dir == 0`); accel-sorted tables go through the comparison branch (generalised to the single accel column) and are positioned, not appended.
- Ties break on `message_id` in the primary sort direction, matching the `(folder_id, <ts>, message_id)` index. The accel load queries and the flat `stbl` readback get the same `message_id` tiebreak so a reload reproduces the order incremental notifications build.

### DIFF
--- a/exch/exmdb/db_engine.cpp
+++ b/exch/exmdb/db_engine.cpp
@@ -521,7 +521,8 @@ table_node::table_node(const table_node &o, clone_t) :
 	folder_id(o.folder_id), handle_guid(o.handle_guid),
 	prestriction(o.prestriction), psorts(o.psorts),
 	instance_tag(o.instance_tag), extremum_tag(o.extremum_tag),
-	header_id(o.header_id), b_search(o.b_search), b_hint(o.b_hint)
+	header_id(o.header_id), accel_tag(o.accel_tag), accel_dir(o.accel_dir),
+	b_search(o.b_search), b_hint(o.b_hint)
 {}
 
 table_node::~table_node()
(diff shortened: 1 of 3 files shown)


jengelh:
:+1:

6db7bfe683f2d11c4abcea00e6def26e0014a187 (some comment re-flowing, whitespacing, more issue numbers)

PR #286, opened 2026-07-22
ews: implement ReplyToItem/ReplyAllToItem/ForwardItem, respect client From on save

nourdineb-ops:
`CreateItem` hard-errored (`E-3045: failed to find proper type for node 'ReplyToItem'`) whenever a client saved a reply or forward as a draft without sending it. Outlook for Mac sends `ReplyToItem`/`ReplyAllToItem`/`ForwardItem` (EWS's "smart response" types) specifically for that action. A direct send, or a brand-new non-reply draft, both use the already-supported plain `tMessage` path instead - that's why replying or forwarding normally seemed to work; only "reply, then keep as draft" was broken.

The wire format Outlook for Mac actually sends nests a full `<t:Message>` element inside `<t:ReplyToItem>`, alongside `<t:ReferenceItemId>`. This differs from `Types.xsd`'s documented `SmartResponseType`, which is flat (`Subject`/`Body`/`ToRecipients`/`From` directly on the element). I captured the actual shape from a live client request during investigation. `tReplyToItem`/`tReplyAllToItem`/`tForwardItem` model that observed shape - a reference id plus a wrapped `tMessage` - rather than the documented schema.

These wrapper types are request-only: they never appear as a stored or loaded item, so they don't fit `sItem`'s generic load/update/serialize machinery (shared by `GetItem`, `UpdateItem`, `MarkAsJunk`, `SyncFolderItems`, etc). I introduced a separate `sCreateItem` variant - `sItem`'s alternatives plus the three wrapper types - scoped to `CreateItem`'s request body only. `sItem` and everything built on it is untouched.

Second commit: a related bug found while testing against a mailbox with a secondary alias domain. `CreateItem`'s `SaveOnly` path for a plain message unconditionally forced the sender identity to the mailbox owner's default address. It discarded any client-supplied "From" - already parsed into `PR_SENT_REPRESENTING_*` by `toContent()`, but never checked before being overwritten. A reply saved from a secondary alias address always reverted to the primary identity. Now it only falls back to the default identity if the client didn't already supply one, matching grommunio-web's own `PR_SENT_REPRESENTING_* ??= PR_SENDER_*` pattern (`class.operations.php`) - applied here to the new wrapper types as well as plain `tMessage`.

Tested against a real gromox-http instance: replayed the exact captured Outlook Mac request (`ReplyToItem` with a nested `Message` and `ReferenceItemId`), confirmed `ResponseClass=Success` with correct MAPI property storage (`PR_SUBJECT_PREFIX`/`PR_NORMALIZED_SUBJECT` split, sender fallback). Same result for `ForwardItem`. Also confirmed live against a real Outlook for Mac client reproducing the original failure (reply to a message, save as draft without sending) - zero errors in gromox-http's log.

### DIFF
--- a/exch/ews/context.cpp
+++ b/exch/ews/context.cpp
@@ -3762,7 +3762,7 @@ EWSContext::MCONT_PTR EWSContext::toContent(const std::string& dir, std::string&
  *
  * @return Pointer to MESSAGE_CONTENT structure
  */
-EWSContext::MCONT_PTR EWSContext::toContent(const std::string& dir, const sFolderSpec& parent, sItem& item, bool persist) const
+EWSContext::MCONT_PTR EWSContext::toContent(const std::string& dir, const sFolderSpec& parent, sCreateItem& item, bool persist) const
 {
 	const auto& exmdb = m_plugin.exmdb;
 	uint64_t messageId, changeNumber;
@@ -4763,6 +4763,31 @@ void EWSContext::toContent(const std::string &dir, tDeclineItem &item,
 	if (!item.ItemClass)
 		item.ItemClass.emplace("IPM.Schedule.Meeting.Resp.Neg");
 	toContent(dir, static_cast<tMessage &>(item), shape, content);
+}
+
+/*
+ * ReplyToItem/ReplyAllToItem/ForwardItem wrap a full <t:Message> rather than
+ * extending tMessage directly (see the comment on tReplyToItem). The message
+ * content itself is otherwise handled exactly like a normal saved message.
+ * ReferenceItemId (the item being replied to/forwarded) is not otherwise acted
+ * upon yet.
+ */
+void EWSContext::toContent(const std::string &dir, tReplyToItem &item,
+    sShape &shape, MCONT_PTR &content) const
+{
+	toContent(dir, item.Message, shape, content);
+}
+
+void EWSContext::toContent(const std::string &dir, tReplyAllToItem &item,
+    sShape &shape, MCONT_PTR &content) const
+{
+	toContent(dir, item.Message, shape, content);
+}
+
+void EWSContext::toContent(const std::string &dir, tForwardItem &item,
+    sShape &shape, MCONT_PTR &content) const
+{
+	toContent(dir, item.Message, shape, content);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


jengelh:
I like this mantra from the [qemu submission guideline](https://www.qemu.org/docs/master/devel/submitting-a-patch.html#split-up-long-patches):

>A commit message that mentions “Also, …” is often a good candidate for splitting into multiple patches.

That seems to be the case here. You should also fix your $GIT_AUTHOR_NAME/$GIT_AUTHOR_EMAIL to be less ambiguous (root@localhost who?).

---

<details><summary>History note</summary>

<p>for the record: that was added to the qemu wiki [16:22Z, 30 June 2014](https://wiki.qemu.org/index.php?title=Contribute/SubmitAPatch&oldid=4211)‎ by Eblake (RH).</p>

</details>

nourdineb-ops:
Good catch on both counts - split into two commits (the ReplyToItem/ReplyAllToItem/ForwardItem implementation, and the separate From-respecting fix), and fixed the author identity for this and future pushes.

jengelh:
>Outlook for Mac uses this "smart response" wrapper

"this smart response wrapper" refersto what? `SmartResponseType`, `sCreateItem`, to a sequence of EWS requests?

In the sentence structures, try to use fewer nested subclauses—especially long inserts (and different ways of doing so)—and yes, this is an in-situ negative example. EWS is complicated as it is, let's at least make the description easier :sweat_smile: 

```diff
+               bool persist = !(isMessageLike && request.MessageDisposition == Enum::SendOnly);
+
                auto content = ctx.toContent(dir, *targetFolder, item, persist);
 
+               if (isMessageLike &&
+                   request.MessageDisposition == Enum::SaveOnly) {
+                  const char *sender = ctx.effectiveUser(*targetFolder);
+                  if (sender == nullptr || *sender == '\0')
+                     sender = ctx.auth_info().username;
+
+                  if (sender != nullptr && *sender != '\0') {
```

Indent is inconsistent here (3-spaced; should be just tab).

nourdineb-ops:
Fair point, rewrote the description - shorter sentences, and "this wrapper" now spells out ReplyToItem/ReplyAllToItem/ForwardItem explicitly instead of relying on the reader tracking the referent back. Thanks for the nudge.

PR #294, opened 2026-07-23
ews: signal ConnectionStatus=Closed when GetStreamingEvents faults a subscription

nourdineb-ops:
A subscription's event backlog can exceed `ews_max_pending_events`. When that happens the server faults it (`EWSPlugin::event()` sets `mgr->overflow`) so that the client re-subscribes and resyncs. `GetStreamingEvents` already reports `ErrorInvalidSubscription` for a faulted subscription, but it unconditionally set `ConnectionStatus=OK` on every response, error or not.

A client that reacts to connection state rather than polling the error code indefinitely gets no signal that this particular streaming connection is done for good. I hit this in production: a real Outlook for Mac client kept calling `GetStreamingEvents` against a faulted subscription every ~50 seconds for over two hours, always getting `ErrorInvalidSubscription` with `ConnectionStatus=OK`, and never re-subscribed on its own. New mail stopped surfacing in the client for the whole window, even though the server was answering every poll with `200 OK`.

This reports `ConnectionStatus=Closed` when the response carries an error, matching the field's documented meaning (the connection is closed, reconnect if you want more events) instead of leaving it at `OK` alongside a fatal error.

Tested by building against the reproduction above (rolled the fix out to the same mailbox that got stuck) - can't easily automate "wait 2+ hours for the real client to give up," but the change itself is a 2-line, low-risk behavioral fix scoped to the already-error branch.

### DIFF
--- a/exch/ews/ews.cpp
+++ b/exch/ews/ews.cpp
@@ -734,12 +734,22 @@ int EWSContext::notify()
 	}
 	for (const tSubscriptionId &subscription : msg.ErrorSubscriptionIds)
 		std::erase(nctx.nct_subs, subscription);
-	msg.success();
+	if (msg.ErrorSubscriptionIds.empty())
+		msg.success();
+	else
+		msg.error("ErrorInvalidSubscription", "Subscription is invalid.");
 	// If there are no more subscriptions to monitor or the stream expired, close it
 	// If there were more events than we could deliver in one message, proceed with the next chunk right away
 	// Otherwise just go back to sleep
 	nctx.state = nctx.nct_subs.empty() || tp_now() > nctx.expire ?
 	             NS::S_CLOSING : moreAny ? NS::S_SLEEP : NS::S_WRITE;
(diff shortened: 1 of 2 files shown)


nourdineb-ops:
Added a follow-up commit extending the same fix to the streaming notify() loop: when an already-linked GetStreamingEvents subscription faults mid-stream (not just on the initial request), the response reported ResponseClass="Success" with the error buried in ErrorSubscriptionIds and never set ConnectionStatus on that chunk - the Closed signal only appeared one chunk later in the unrelated close message, disconnected from the error detail. This commit applies the same success()/error() choice already used in the request handler and sets ConnectionStatus consistently with the computed state transition, so a client reading a single chunk gets a complete, correct signal.

jengelh (commented):
>Observed in production: a real Outlook for Mac client kept calling

What's the deal here that necessitates putting emphasis on OFM being "real"?

PR #296, opened 2026-07-27
ews: personal Contacts folder integration for GetUserPhoto and FindPeople

nourdineb-ops:
Stacked on top of #270 (adds tPersona::PersonaId/RelevanceScore/GivenName/Surname this depends on, and findpeople_search_recipient_history() this is placed alongside) - GitHub cannot base a PR against another fork's branch, so this diff includes #270's two commits too; only the last two commits (`ews: fall back to a personal Contacts photo in GetUserPhoto` and `ews: search the requesting user's own Contacts folder in FindPeople`) are new here. Please review/merge #270 first, or let me know if you'd rather review these together.

Two related fixes making EWS aware of the requesting user's own personal Contacts folder, which neither GetUserPhoto nor FindPeople looked at before:

- `EWSContext::findContactPhoto()`: GetUserPhoto only ever served a local mailbox user's own organizational profile photo (a store-level named property). It had no notion of a personal contact's own photo (a hidden PR_ATTACHMENT_CONTACTPHOTO attachment on an IPM.Contact item) - the far more common real-world source of a photo for an external address someone has simply saved as a contact. This adds it as a fallback, scoped to the requesting user's own Contacts folder only.

- `findpeople_search_contacts()`: FindPeople only ever searched the ab_tree/GAL and grommunio-web's own recipient-history JSON blob (people the user has sent mail to via the webapp specifically). A real personal contact, saved with full details but never independently emailed through the webapp, was invisible to Outlook's autocomplete no matter how complete the contact card was. This adds a direct search of the requesting user's own Contacts folder, matching PR_DISPLAY_NAME or any of Email1/2/3EmailAddress by substring - the same restriction shape grommunio-web's own ResolveNamesModule::searchContactsFolders() uses for the equivalent lookup.

Both deployed to production and confirmed working live (contact photo shown in message list/reading pane; a personal contact never emailed via the webapp now correctly autocompletes in real Outlook).

### DIFF
--- a/exch/ews/context.cpp
+++ b/exch/ews/context.cpp
@@ -3332,6 +3332,74 @@ sFolderSpec EWSContext::resolveFolder(const sMessageEntryId& eid) const
 }
 
 /**
+ * @brief      Find a contact photo by email address in the given store's
+ *             default Contacts folder
+ *
+ * @param      dir    Home directory of the mailbox whose Contacts folder to search
+ * @param      email  Email address to match against Email1/2/3
+ *
+ * @return     Pointer to the photo's binary data, or nullptr if no matching
+ *             contact (or no photo on it) was found
+ *
+ * @note       Only searches the default Contacts folder, not any user-created
+ *             subfolders under it.
+ */
+const BINARY *EWSContext::findContactPhoto(const std::string &dir,
+    const std::string &email) const
+{
+	std::array<PROPERTY_NAME, 3> names = {NtEmailAddress1, NtEmailAddress2, NtEmailAddress3};
+	PROPNAME_ARRAY propNames = {static_cast<uint16_t>(names.size()), names.data()};
+	PROPID_ARRAY propIds = getNamedPropIds(dir, propNames);
+	if (propIds.size() != names.size())
+		return nullptr;
+
+	std::array<RESTRICTION_PROPERTY, 3> propertyRestrictions{};
+	std::array<RESTRICTION, 3> childRestrictions{};
+	TAGGED_PROPVAL matchValue{0, deconst(email.c_str())};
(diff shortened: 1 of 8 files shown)


nourdineb-ops:
Friendly ping on this one - no rush, just making sure it did not fall through the cracks. Happy to add tests or rework anything that does not fit the codebase conventions if that's what's blocking a look.

jengelh:
I am thinking we need to hoist the `case PR_SMTP_ADDRESS:` from nsp and zcore/ab_tree.cpp (in)to `ab_node::fetch_prop`. (The addressbook should respond the same for any component querying it.) Other properties may need the same treatment.

nourdineb-ops:
Good call - done in fb6e3322c. Moved the `mlist_info`/`user_info(mail_address)` dispatch out of nsp_interface.cpp and zcore/ab_tree.cpp and into `ab_base::fetch_prop()`, both call sites now just do `node.fetch_prop(PR_SMTP_ADDRESS, dn)`.

Turns out this directly benefits this PR too: the FindPeople/GetUserPhoto code here had its own workaround for the exact same gap (`node.user_info(...)` instead of `node.fetch_prop(PR_SMTP_ADDRESS, ...)`, with a comment explaining why fetch_prop never worked for that tag) - removed both now that fetch_prop() actually resolves it.

Didn't touch PR_ACCOUNT/PR_ACCOUNT_A/PR_SMTP_ADDRESS_A - nsp_interface.cpp maps all four to the same value today, so they're covered transparently through the same case block, no separate handling needed. Tested live: NSP-backed lookups and EWS FindPeople both still resolve the same addresses as before, no behavior change for existing callers.

PR #298, opened 2026-07-28
mra: coverity issue "inefficient resizing" fixed

EugenePopovych:
To avoid coverity issue "Inefficient vector resizing with reserve."
we hide the actual logic behind std::min call.


jengelh:
This needs std::max though.

PR #299, opened 2026-07-28
exch: fixed "using moved object" coverity issue

EugenePopovych:


jengelh:
There is a line

```c++
par->conn = std::make_shared<exmdb_connection>(std::move(co));
```

which will transfer the `co` object irrespective of whether `exmdb_parser_insert_conn` receives it by lvalue-ref or rvalue-ref, i.e. conn.sockd is still unset after this change.

PR #303, opened 2026-08-04
ews: add conversationhistory, recipientcache and archive folders

codessensei:
### Problem

When Outlook for Mac (16.111) sets up an account it resolves a batch of
distinguished folder ids in a single `GetFolder` call. Three of the names are
unknown to Gromox and answered with `ErrorFolderNotFound`:

```
E-3051: unknown distinguished folder id 'conversationhistory'
E-3051: unknown distinguished folder id 'recipientcache'
E-3051: unknown distinguished folder id 'archive'
```

The client then repeats the identical batch every 15 seconds and never
advances to `SyncFolderHierarchy` / `SyncFolderItems`. The account
authenticates and stays connected, but only the locally created Inbox and
Drafts are shown and no mail is ever fetched.

The server side is otherwise fine: issuing `FindItem`, `SyncFolderItems` and
`SyncFolderHierarchy` by hand against the same mailbox returns the messages and
all 19 folders.

### Change

Add the three folders, using the id range already reserved by
`PRIVATE_FID_UNASSIGNED_START`:

| name | id | class | hidden |
|---|---|---|---|
| `conversationhistory` | 0x1e | IPF.Note | no |
| `recipientcache` | 0x1f | IPF.Contact | yes |
| `archive` | 0x20 | IPF.Note | no |

`data/folder_names.txt` already contained an `Archive` display name and a
`ConversationHistory` translation for every locale, so only the numeric ids
needed to be added to the `[en]` fallback block.

`std::array<DistNameInfo, N>` in `structures.hpp` is a fixed size and had to
grow with the table; the list is kept alphabetically sorted as its comment
requires.

### Result

The `GetFolder` batch Outlook sends goes from 12/16 to 15/16 resolved, and the
account synchronises normally — folders, messages and attachments all appear.

### Open question: `archivemsgfolderroot`

The remaining unresolved name is `archivemsgfolderroot`, which in Exchange
refers to the root of a *separate* archive mailbox. Gromox has no archive
mailbox concept, so `ErrorFolderNotFound` is arguably the correct answer — real
Exchange returns the same when no archive is provisioned. Outlook for Mac
appears not to tolerate it, though.

Aliasing it to the primary `Archive` folder makes the client happy but is
semantically wrong, so it is deliberately left out of this PR. I am happy to
follow up with whichever approach you prefer.

### Existing stores

`mkprivate` creates the folders for new stores only. Stores created before this
change do not get them and would need a schema upgrade step; that is not part
of this PR.

Built and tested on Debian 13.


jengelh:
Duplicate of https://github.com/grommunio/gromox/pull/273

PR #306, opened 2026-08-05
cov int overflow fix

EugenePopovych:
- **exch: fixed "using moved object" coverity issue**
- **Fix coverity issue "Possible integer overflow"**


jengelh:
I have split this one commit up into three.

PR #307, opened 2026-08-05
emsmdb,zcore: retain sent-item copies for shared mailboxes (MessageCopyForSentAs)

r4nc0r:
## What this is for

When someone sends a message as a shared mailbox, or on behalf of it, only that person's own Sent Items keeps a record. Nobody else with access to the shared mailbox can see what went out in its name, which is usually the whole point of having it shared. Exchange solves this with two mailbox settings, `MessageCopyForSentAsEnabled` and `MessageCopyForSendOnBehalfEnabled`; Gromox had no equivalent.

With this change, a mailbox can ask for a copy of such messages to be filed into its own Sent Items folder. The person who sent it still keeps their own copy, which is what Exchange does too.

## How an admin turns it on

```
gromox-mbop -u shared@example.com set-msgcopy -a 1   # sending as
gromox-mbop -u shared@example.com set-msgcopy -b 1   # sending on behalf of
gromox-mbop -u shared@example.com get-msgcopy        # show both
gromox-mbop -u shared@example.com clear-msgcopy      # back to unset
```

The settings live in the represented mailbox's own store, as two `PT_BOOLEAN` named properties under `PSETID_Gromox` (`msgcopy_sentas` and `msgcopy_sendonbehalf`), the same way `websettings` and `photo` are held, and are read at submit time from the mailbox being represented — not from the account doing the sending.

## How it is wired

The copy is deposited from `cu_send_message`, in both the emsmdb and the zcore implementation, right after the message has gone out over SMTP and before the `PR_TARGET_ENTRYID` and `PR_DELETE_AFTER_SUBMIT` branches, which both return early: what the client wants done with the sender's own copy should not decide anything about the represented mailbox's copy.

`cu_send_message` previously took a ready-made Envelope-From string. It now takes the actor, the delegator and the grant level, and derives Envelope-From itself, which also removes the identical conditional that all three call sites were repeating. `repr_grant` was defined twice, identically, in `oxomsg.cpp` and in zcore's `common_util.hpp`; it moved to `mapidefs.h` next to the `frights*` bits it is derived from.

Turning a just-submitted message into something that can be written into a different store is shared code in `lib/mapi/sent_copy.cpp`. It remaps named property ids, because those are assigned per store, and drops the properties that name the source store or describe the submission that just finished. The named-property resolvers are passed in rather than called directly, since emsmdb and zcore each have an exmdb client object of a different type; the callback signatures mirror the EXMIDL prototypes so both can be assigned without an adapter.

## Three details that are deliberate

A property the target store will not name is **dropped from the copy** rather than carried over. `get_named_propids` reports id 0 for names it did not create — including the case where it silently stops creating because the target store reached `MAXIMUM_PROPNAME_NUMBER` — while still succeeding overall. Keeping the source id there would not preserve the property, it would name an unrelated property in the target store. Dropped values are counted and logged rather than lost quietly.

The guard is on the **grant level**, not on a delegator being named. `IPM.Schedule` messages are let through without any delegation check at all (`pass_scheduling`), so an unauthorised third party can appear as the delegator on that path.

The "am I representing myself" test compares the **canonical username and the maildir**, not the address that was requested. Sending under an alias of one's own account is also reported as `send_as`, so comparing addresses would file an ordinary aliased send into the sender's own Sent Items twice.

## Failure handling

Nothing here can turn a completed send into a failure. The mail has already left the system when the copy is attempted, so every failure is logged and dropped, and the hook has its own `bad_alloc` handler rather than falling through to the enclosing one, which answers `ecServerOOM`. A mailbox that is over quota, unreachable, or unreadable therefore costs a log line and no copy. The folder ACL is read but not enforced, matching Exchange, because the delegate lists and folder permissions are administered separately and a delegate without an explicit create right on Sent Items is an ordinary configuration; writing without a confirmed right leaves a trace.

## Not covered

EWS does not honour these settings. `EWSContext::send` performs no delegation check at all and always uses the authenticated user as Envelope-From, so Send-As has to exist there before a copy could be derived from it. Deferred sends return before this code and complete through `submit_command`, and whether a copy results on that path has not been established. Both are recorded in `doc/known_bugs.rst`.


include/gromox/mapitags.hpp
```diff
@@ -884,6 +884,8 @@ enum {
 	PR_EC_EXTERNAL_AUDIENCE = PROP_TAG(PT_BOOLEAN, 0x6766),
 	PR_EC_EXTERNAL_REPLY = PROP_TAG(PT_UNICODE, 0x6767),
 	PR_EC_EXTERNAL_SUBJECT = PROP_TAG(PT_UNICODE, 0x6768),
+	PR_MESSAGE_COPY_FOR_SENT_AS = PROP_TAG(PT_BOOLEAN, 0x6769), /* Gromox-specific */
```
jengelh: I would rather like that new properties are Named Properties (similar to `PSETID_Gromox`:`websettings`). Clogging up the 0x6600..0x67ff space like Kopano did is not overly scalable (less than 512 props).

r4nc0r:
Done — pushed as a separate commit on top so the change is easy to read against what you reviewed.

The two settings are now `MNID_STRING` named properties under `PSETID_Gromox`, named `msgcopy_sentas` and `msgcopy_sendonbehalf`, held on the store exactly as `websettings` and `photo` are. `PR_MESSAGE_COPY_FOR_SENT_AS` and `PR_MESSAGE_COPY_FOR_SEND_ON_BEHALF` are gone, so `include/gromox/mapitags.hpp` is untouched by this PR now. The names are one line each to change if you would rather have them spelled differently.

Three things worth flagging about the conversion:

`logon_object`'s resolver was not usable on the send path. It answers for the logged-on store and caches into it, and the store that carries the setting is the delegator's, so it would have queried the wrong mailbox and poisoned the cache. `exmdb_client->get_named_propid(dir, ...)` is used instead, with the zcore free-function equivalent in the zcore twin. zcore's `cu_read_storenamedprop` was not usable either, since it reports a failed lookup and an unset property alike, and keeping those apart is what the separate warning on the read exists for.

The read resolves with `b_create` unset, so sending in a mailbox's name never brings the property into existence there. A propid of 0 on the read path is simply "never configured" and returns silently. Resolution also happens after the "am I representing myself" guard, so an ordinary send still reaches none of this — the extra RPC is only paid on a genuinely delegated send.

`set-msgcopy` resolves with `b_create` set and now fails loudly if it gets 0 back. That can only mean the store has reached `MAXIMUM_PROPNAME_NUMBER`, where `get_named_propids` stops minting names and reports 0 while still returning success; without the check an operator would be told the setting was applied when it was not. `get-msgcopy` and `clear-msgcopy` treat an unresolvable name as the property being unset, which is the state they would report and produce anyway.

Build is clean. As before, this is compile-verified only and has not been exercised at runtime — the named-property resolution included.

r4nc0r:
this would also be a huge help for our organisation, we have multiple shared mailboxes that are used via Outlook. And someone answers an email from the shared mailbox the others cannot see the answer :(

also #317 is blocking us from using my Contributions

r4nc0r:
@jengelh could you review this again? Our Outlook useres cannot work on shared mailboxes without this

dneuhaeuser:
Perhaps this could be expanded to also include a setting where sent emails are saved ONLY to the “Sent” folder of the SHARED mailbox.

I have use-cases where users have 2 personal mailboxes and they want to see each sent message only once.

Exchange behaviour is described here:
https://learn.microsoft.com/en-us/troubleshoot/exchange/user-and-shared-mailboxes/sent-mail-is-not-saved
There's the Outlook Client Setting "DelegateSentItemsStyle" and the Server-side Setting "MessageCopyForSentAsEnabled".
**with DelegateSentItemsStyle=1 and MessageCopyForSentAsEnabled=False 
-> A copy of the email will be saved in the shared mailbox and **no** copies in the primary mailbox.**

In Grommunio, with DelegateSentItemsStyle=1, currently sent messages are saved in BOTH mbx, primary and shared.

r4nc0r:
Could you give me your Outlook version and the RegKey for DelegateSentItemsStyle=1?
Our Outlook ignores this setting :(

I'll look into extending this to so your use case is also covered.

exch/emsmdb/common_util.cpp
```diff
@@ -1385,11 +1386,121 @@ static BOOL common_util_get_propname(propid_t propid, PROPERTY_NAME **pppropname
...
+	 * Only leave a trace of having written without a confirmed one — which
+	 * covers the permission being absent and the lookup having failed
+	 * alike, so that the trace does not vanish in the degraded case.
+	 */
+	uint32_t perm = 0;
+	if (!exmdb_client->get_folder_perm(mres.maildir.c_str(), fid, actor,
+	    &perm) || !(perm & (frightsOwner | frightsCreate)))
+		mlog(LV_INFO, "I-1283: depositing a copy of %s in <%s>'s Sent "
+			"Items without a confirmed create right for %s",
+			log_id, delegator, actor);
```
jengelh, line 1494: I wonder what the situation is when the actor is also store owner of the delegate (STORE_OWNER_GRANTED-style mechanism), which I don't think get_folder_perm would report.
Leaving an extra commit without fixing is ok.

exch/emsmdb/common_util.cpp
```diff
@@ -1385,11 +1386,121 @@ static BOOL common_util_get_propname(propid_t propid, PROPERTY_NAME **pppropname
...
+	if (mysql_adaptor_meta(delegator, WANTPRIV_METAONLY, mres) != 0) {
+		mlog(LV_WARN, "W-1273: <%s> cannot be resolved, so no copy of "
+			"%s was deposited", delegator, log_id);
+		return;
+	}
+	auto dir = plogon->get_dir();
+	/*
+	 * Sending under an alias of one's own account is also reported as
+	 * send_as, and there @delegator differs from @actor while the mailbox is
+	 * the same one; filing that would duplicate the sender's own copy. The
```
jengelh: Really not a fan of `:`, `;` and `—`.

doc/gromox-mbop.8
```diff
@@ -419,6 +425,67 @@ To process an entire mailbox and wipe everything older than a few days:
...
+appears in the log. The quota of the represented mailbox is not consulted
+beforehand, and the copy counts against it like any other message.
+.PP
+Two paths do not produce a copy at this time. \fBThe EWS protocol\fP does not,
+so messages sent by clients using EWS are unaffected by these settings.
+\fBDeferred sends\fP (a client asking for delayed delivery) complete through a
+separate helper, and whether a copy results there has not been established.
+.SS Options
+.TP
+\fB\-a\fP [\fIboolean\fP]
```
jengelh: the boolean isn't optional so should not be enclosed in `[ ]`.

doc/gromox-mbop.8
```diff
@@ -419,6 +425,67 @@ To process an entire mailbox and wipe everything older than a few days:
...
+.SS Synopsis
+\fBget\-msgcopy\fP
+.PP
+\fBset\-msgcopy\fP [\fB\-v\fP] [\fB\-a\fP \fIboolean\fP] [\fB\-b\fP \fIboolean\fP]
+.PP
+\fBclear\-msgcopy\fP [\fB\-abv\fP]
+.SS Description
+Normally, a message which one user sends as, or on behalf of, another mailbox
+is filed only in the Sent Items folder of whoever composed it, and the
+represented mailbox retains no record of it. This is inconvenient for shared
```
jengelh: I find this a really long-winded explanation. It's a manual, it should explain what the commands do, not how email works.

doc/known_bugs.rst
```diff
@@ -24,6 +25,23 @@ Noteworthy issues for administrators and packagers
...
 
+Gaps in Gromox
+==============
+
+* The MessageCopyForSentAs and MessageCopyForSendOnBehalf mailbox settings (see
+  gromox-mbop(8)) are evaluated by emsmdb and zcore only. Messages submitted
+  over **EWS** never produce a copy in the represented mailbox, because
+  ``EWSContext::send`` performs no delegation check at all and always uses the
+  authenticated user as Envelope-From; Send-As has to exist there before a copy
+  can be derived from it.
```
jengelh: In case you have not seen it, there is a concurrent submission https://github.com/grommunio/gromox/pull/316 . Worth taking in first?

tools/mbop_msgcopy.cpp
```diff
@@ -0,0 +1,198 @@
...
+	MBOP_AUTOHELP,
+	HXOPT_TABLEEND,
+};
+
+/**
+ * Unlike parse_bool, an unrecognized word is rejected rather than taken to mean
+ * "true". Enabling this by way of a typo would silently start copying mail into
+ * someone else's mailbox.
+ */
+static bool parse_strict_bool(const char *s, uint8_t *out)
```
jengelh: Sounds like a something that would be useful elsewhere. Let's move this func to lib/string.cpp.

doc/gromox-mbop.8
```diff
@@ -456,6 +456,11 @@ get\-msgcopy reports the current state of both settings. A setting which was
 never set is reported as such, and behaves as if switched off; clear\-msgcopy
 returns a setting to that state, as opposed to setting it to a recorded 0.
 .PP
+The settings are held as named properties of the mailbox's store, in the
```
jengelh: That's a MAPI detail and hardly ever relevant for admins. If going by Exchange standards, 32k named props is enough for everyone. If the situation occurs, the program still errors out, but it's not worth bloating the doc over.

exch/emsmdb/common_util.cpp
```diff
@@ -1433,6 +1430,27 @@ static void cu_deposit_repr_copy(logon_object *plogon,
 	 */
 	if (strcasecmp(mres.username.c_str(), actor) == 0 || mres.maildir == dir)
 		return;
+	/*
+	 * b_create=FALSE: a mailbox which never had the setting written must not
+	 * acquire a named-property row from every message sent in its name.
```
jengelh: Fold the change from PR_EC_ to namedprop into the earlier commits.

jengelh (changes requested):
general upvote

exch/emsmdb/common_util.cpp
r4nc0r: You were right. `frightsOwner` on `PRIVATE_FID_IPMSUBTREE` is what Gromox treats as store ownership, `get_mbox_perm` derives `frightsGromoxStoreOwner` from it, and `cu_get_folder_permission` never reports it because it only reads the `permissions` table. So a delegate who holds store ownership on the shared mailbox was getting the `I-1283` line every time. Fixed in its own commit, as you allowed. `get_mbox_perm` is only asked when the folder ACL grants nothing, so a mailbox with an ordinary ACE still costs one RPC.

doc/known_bugs.rst
r4nc0r: Had not seen it, thanks. Yes, take it first. It is the piece the `known_bugs.rst` entry here names as missing, since without a Send-As check on the EWS side there is no authorization to derive a copy from. This PR does not depend on it, different base branch, and the entry is worded to stay accurate either way.

r4nc0r:
Thanks, all eight are addressed. Rebased onto current prime while I was in there.

@dneuhaeuser 

Implemented, thanks for the pointer to the Exchange behaviour.

There is now a third setting, `exclusive`, on the same mailbox and set the same way:

```
gromox-mbop -u shared@example.com set-msgcopy -a 1 -x 1
```

With it, the sender's own copy is discarded instead of being filed anywhere in their own mailbox, so the shared mailbox holds the only copy. That is the `DelegateSentItemsStyle=1` plus `MessageCopyForSentAsEnabled=False` outcome from the article, reached from the server side alone, which also means it does not depend on the client honouring the registry value.

Three details that matter for your two-mailbox case:

It has no effect on its own. It only says what happens to the sender's copy once `sent-as` or `send-on-behalf` has produced a copy in the shared mailbox, so `-x 1` by itself changes nothing.

It is only acted on after the shared mailbox's copy is safely written. If that mailbox is over quota or unreachable, the sender keeps their copy and you get a warning in the log. A sent message cannot end up filed nowhere.

It takes effect ahead of `PR_TARGET_ENTRYID`, so a client that files its own copy into the shared mailbox does not put a second one back.

Still open on my side is your `DelegateSentItemsStyle` question. Our Outlook ignores that value, so if you can send me your Outlook version and the exact registry key and path you set, I would like to work out why. It does not block anything here, since the new setting produces the same end state without the client's help, but I would rather understand it.

r4nc0r:
A delegated send through zcore, one account sending as a shared mailbox listed in its `config/sendas.txt`, counted in both mailboxes after each send.

| `sent-as` | `exclusive` | Shared mailbox Sent Items | Sender's Sent Items |
|:---|:---|:---|:---|
| off | off | no copy | copy |
| off | **on** | no copy | copy |
| on | off | copy | copy |
| on | **on** | copy | **no copy** |

dneuhaeuser:
> There is now a third setting, exclusive, on the same mailbox and set the same way...

Thank you very much for including this. Looking forward to test the new feature.


> Could you give me your Outlook version and the RegKey for DelegateSentItemsStyle=1? Our Outlook ignores this setting :(

Sure, the RegKey looks like this:
```
[HKEY_CURRENT_USER\Software\Microsoft\Office\16.0\Outlook\Preferences]
"DelegateSentItemsStyle"=dword:00000001
```
This should work with Outlook versions 2016-2024 and 365 as well.
Outlook profile needs to use cache-mode for that!

r4nc0r:
ahh, we use online mode. thanks!

include/gromox/sent_copy.hpp
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0-only WITH linking exception
+// SPDX-FileCopyrightText: 2026 grommunio GmbH
+// This file is part of Gromox.
+#pragma once
```
jengelh, line 4: [Based on the sent-as, I have retargeted the PR for master.]

We would prefer if contributions are made available under AGPL-3.0-or-later terms. Can you accomodate this request?

lib/mapi/sent_copy.cpp
```diff
@@ -0,0 +1,251 @@
...
+	std::vector<proptag_t> drop;
+	for (const auto &pv : props) {
+		if (!is_nameprop_id(PROP_ID(pv.proptag)))
+			continue;
+		auto it = std::find(src.begin(), src.end(), PROP_ID(pv.proptag));
+		if (it != src.end() && dst[it-src.begin()] == 0)
+			drop.push_back(pv.proptag);
+	}
+	for (auto tag : drop)
+		props.erase(tag);
```
jengelh: Hm, two iterations. Perhaps this could benefit from std::erase_if?

```
auto dropped = std::erase_if(props, [&](TAGGED_PROPVAL &pv) {
    if (!is_nameprop_id(PROP_ID(pv.proptag)))
        return false;
    auto it = std::find(src.begin(), src.end(), PROP_ID(pv.proptag));
    return it != src.end() && dst[it-src.begin()] == 0;
};
...
return dropped;
```

lib/mapi/sent_copy.cpp
```diff
@@ -0,0 +1,251 @@
...
+			continue;
+		auto it = std::find(src.begin(), src.end(), PROP_ID(oldtag));
+		if (it == src.end())
+			continue;
+		auto newid = dst[it-src.begin()];
+		if (newid == 0)
+			continue; /* already removed above */
+		pv.proptag = PROP_TAG(PROP_TYPE(oldtag), newid);
+	}
+	return drop.size();
```
jengelh: Hm, three iterations. Perhaps this could be compacted to just one, editing props as it goes and counting drops in a `size_t dropped` without building a dedicated vector just for it.

lib/mapi/sent_copy.cpp
```diff
@@ -0,0 +1,251 @@
...
+	 * message as an object in a store. An embedded message is stored inside
+	 * its attachment rather than as an addressable message of its own, so
+	 * its copies of these carry no meaning to strip.
+	 */
+	for (auto tag : strip_location)
+		dst->proplist.erase(tag);
+	for (auto tag : strip_recomputed)
+		dst->proplist.erase(tag);
+	for (auto tag : strip_submit)
+		dst->proplist.erase(tag);
```
jengelh, line 208: The undesired props could be stripped before npid_remap. (A little less work for npid_remap to jump over non-named props.)

exch/zcore/common_util.cpp
```diff
@@ -164,13 +165,12 @@ static int cu_test_delegate_perm_MD(const char *account,
...
 	for (const auto &d : delegate_list)
-		if (strcasecmp(d.c_str(), account) == 0 ||
-		    mysql_adaptor_check_mlist_include(d.c_str(), account))
+		if (strcasecmp(d.c_str(), account) == 0)
 			return 1;
 	return 0;
 } catch (const std::bad_alloc &) {
 	mlog(LV_ERR, "%s: ENOMEM", __func__);
-	return -1;
+	return false;
```
jengelh: The function is returning `int`. Returning `false` now seems logically incorrect. Should it be `0` or something?

exch/zcore/zserver.cpp
```diff
@@ -125,28 +125,8 @@ USER_INFO_REF zs_query_session(GUID hsession)
...
-		mlog(LV_WARN, "zs_query_session: timeout waiting for session "
-			"lock of user %d; another request is stuck on this session",
-			user_id);
-		tl_hold.lock();
-		pinfo->reference --;
-		tl_hold.unlock();
-		return nullptr;
-	}
 	g_info_key = pinfo;
+	pinfo->lock.lock();
```
jengelh: accidental removal caused by rebase?

exch/zcore/zserver.cpp
```diff
@@ -2853,7 +2833,10 @@ ec_error_t zs_sorttable(GUID hsession,
 		}
 		if (!table_acceptable_type(type))
 			return ecInvalidParam;
-		if (tablesort_is_minmax(psortset->psort[i].table_sort)) {
+		if (TABLE_SORT_MAXIMUM_CATEGORY ==
+			psortset->psort[i].table_sort ||
+			TABLE_SORT_MINIMUM_CATEGORY ==
+			psortset->psort[i].table_sort) {
```
jengelh: accidental reversal due to rebase

jengelh (changes requested):

include/gromox/sent_copy.hpp
r4nc0r: Relicensed.

lib/mapi/sent_copy.cpp
r4nc0r: `npid_apply `is now a single traversal with no intermediate vector. The suggested call does not apply directly: `TPROPVAL_ARRAY `is not a standard container, and its own `erase_if `took a plain function pointer over a const reference — so a predicate could neither consult the id maps nor rewrite a surviving proptag.

That member is now a template over the predicate. The four existing callers in `tools/kdb2mt.cpp` pass a plain function and are unaffected. The ordering constraint the old comment documented disappears with the rewrite, because nothing is matched by proptag any more.

lib/mapi/sent_copy.cpp
r4nc0r: Reordered. Every tag in the three strip lists is a fixed proptag, so nothing `npid_remap `would have acted on is removed early.

jengelh:
please fold the chunks from 93db0b4.

lib/mapi/sent_copy.cpp
r4nc0r: same as above

exch/zcore/common_util.cpp
r4nc0r: The false was not a new line either — it belongs to the same damage. Commit 5e875bbbb reverted six upstream sites; two were caught in review. All six are restored, verified by comparing the set of files containing each affected symbol against upstream/master.

exch/zcore/zserver.cpp
r4nc0r: fixed

exch/zcore/zserver.cpp
r4nc0r: fixed

r4nc0r:
done

PR #311, opened 2026-08-07
ews: make the streaming subscription timeout configurable

nourdineb-ops:
## Summary

`EWSContext::subscribe()` for streaming subscriptions hardcoded the subscription manager's timeout to `5` (minutes) regardless of what's sane for real-world usage:

```c++
return subscribe(req.FolderIds ? *req.FolderIds : std::vector<sFolderId>(),
       req.eventMask(), all, 5);
```

This value backs the cache entry for the subscription (`tSubscriptionId::timeout`, minutes -> `ID.timeout * 60'000` ms in the cache TTL) - once it expires without a `GetStreamingEvents`/`GetEvents` touch, the subscription is evicted server-side.

5 minutes is short enough that ordinary client-side gaps (laptop sleep, travel, network roaming) can outlast it: the subscription disappears silently and the client is left polling a dead id with no way to tell, short of resubscribing from scratch. Observed this contribute to real Outlook desync in production.

## Fix

Added `ews_streaming_subscription_timeout` (default `1440` = 24h) so this is configurable instead of hardcoded. It's bounded in practice by `ews_max_pending_events` regardless of how long it's set to, so a generous default is low-risk.

## Testing

Compiles clean; `libgromox_ews.la` links. Running with the new default in this deployment.

### DIFF
--- a/exch/ews/context.cpp
+++ b/exch/ews/context.cpp
@@ -4911,7 +4911,7 @@ tSubscriptionId EWSContext::subscribe(const tStreamingSubscriptionRequest& req) 
 	if (all && req.FolderIds)
 		throw EWSError::InvalidSubscriptionRequest(E3383);
 	return subscribe(req.FolderIds ? *req.FolderIds : std::vector<sFolderId>(),
-	       req.eventMask(), all, 5);
+	       req.eventMask(), all, m_plugin.streaming_subscription_timeout);
 }
 
 /**
--- a/exch/ews/ews.cpp
+++ b/exch/ews/ews.cpp
@@ -512,6 +512,7 @@ static constexpr cfg_directive ews_cfg_defaults[] = {
 	{"ews_request_logging", "0"},
 	{"ews_response_logging", "0"},
 	{"ews_schema_version", "V2017_07_11"},
+	{"ews_streaming_subscription_timeout", "1440"},
 	CFG_TABLE_END,
 };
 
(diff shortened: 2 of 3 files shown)


jengelh:
>the subscription is silently evicted server-side and the client is left polling a dead id forever, with no way to tell short of resubscribing from scratch.

Can we not respond to dead id poll requests with a reset? The ews/ code should be aware of which subscriptions ids are still alive, so..

nourdineb-ops:
Ran a longer, more rigorous round of empirical testing on this in production since my last comment, and want to correct/refine what I said earlier.

**The reset mechanism from #294 does work, and the client can react to it quickly**: in one test, `ErrorInvalidSubscription`/`ConnectionStatus=Closed` fired on an expired subscription, and MacOutlook issued a fresh `Subscribe` 9 seconds later.

**But that recovery isn't reliable.** In a separate test simulating a real ~1h network outage (subscription alive, then genuinely unreachable, mail delivered mid-outage, then reconnected), Outlook never retried `Subscribe` at all in the time I could observe — it just silently fell back to periodic `SyncFolderItems`/`FindItem` polling to eventually pick up the mail, with push left dead.

**Separately, and probably the more interesting finding**: even on a *healthy* subscription that never errors, I could reproducibly get Outlook to stop acting on new `NewMailEvent`/`DeletedEvent`/etc. notifications after the first one it received following a `Subscribe` — confirmed server-side (the event is generated, queued, and correctly serialized in the response every single time, verified down to the raw bytes on the wire via a temporary mitmproxy capture) and confirmed client-side (no UI change in Outlook despite a valid notification having just been sent). This isn't a signaling problem on our end - the server does its job on every single request - it's Outlook declining to refresh after some point.

Given all of this, gromox has no lever over the client's recovery behavior, which is demonstrably unreliable in at least two independent ways. The practical mitigation available server-side is reducing how often a routine gap (sleep, brief network loss) forces the client into that unreliable recovery path in the first place - which is what this PR does. Happy to go into more detail on any of the above if useful.

jengelh:
If Outlook breaks regardless, then no switching of default timeouts will help. At that point the user just has to bear with Outlook's typical bugs.

Cherry-picked the config directive as 60693128c267c0369ec35994b9b9663fffe1b9a9 .

PR #312, opened 2026-08-10
exmdb: index conversation ids again, with a partial index

r4nc0r:
## What is slow

Clicking a mail in conversation view takes about two seconds on a large mailbox.

When you click a mail there, Outlook asks the server "which mails belong to this same conversation?". The server answers that by looking at message property *values*, across the whole mailbox, because a conversation can have mails in any folder -- so there is no folder and no mail id to narrow the search with. Without an index on those values, the database reads every property row for every click: 47.6 million of them on the store I measured.

## The fix

Index those values again, but only for the one property tag that needs it (`PR_CONVERSATION_ID`). A partial index holds one entry per mail instead of one per property value, which is what made the previous full index expensive.

## Numbers

400k-message store, 47.6M property rows. Times are for the whole conversation-members table load, measured inside exmdb:

| | query plan | table load |
|---|---|---|
| before | `SCAN message_properties` | 1789.8 ms |
| after | `SEARCH message_properties USING INDEX conv_id_index` | 0.3-0.5 ms |

Both confirmed against a real Outlook client clicking mails in a 400k-message folder in conversation view. Running the bare SQL through a CLI agrees on the before figure at 1780 ms; after the change it lands at that tool's floor (3 ms, which is what it also costs for `SELECT 1` against this store), so the CLI cannot resolve the improvement and the in-process figures above are the meaningful ones.

Store size grows by 9 MB of 3166 MB (+0.28%). The full index that was removed at schema level 25 cost about 44%.

Added to the schema for new stores and as upgrade level 27 for existing ones, registered for private and public stores in lockstep. Upgrading the 3.1 GB test store took 2.06 s.

### DIFF
--- a/lib/dbop_sqlite.cpp
+++ b/lib/dbop_sqlite.cpp
@@ -163,14 +163,15 @@ static constexpr char tbl_msgprops_4[] =
 static constexpr char tbl_msgprops_move4[] =
 "INSERT INTO message_properties SELECT message_id, proptag, propval FROM u0";
 
-static constexpr char tbl_msgprops_25[] =
+static constexpr char tbl_msgprops_27[] =
 "CREATE TABLE message_properties ("
 "  message_id INTEGER NOT NULL,"
 "  proptag INTEGER NOT NULL,"
 "  propval BLOB NOT NULL,"
 "  FOREIGN KEY (message_id) REFERENCES messages (message_id) ON DELETE CASCADE ON UPDATE CASCADE);"
 "CREATE INDEX mid_properties_index4 ON message_properties(message_id);"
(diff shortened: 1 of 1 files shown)


r4nc0r:
@jengelh could youtake a look at this our power users use conversation view and working without this fix is a real bummer

jengelh:
>Both confirmed against a real Outlook client clicking mails

What's the deal here that necessitates putting emphasis on Outlook being "real"?

r4nc0r:
i wrote some custom perf scripts to better test/measure the changes. at last i tested it on a "real" outlook client locally against my development gromox

PR #313, opened 2026-08-10
exmdb, emsmdb: make opening a large folder faster

r4nc0r:
## What is slow

Opening a big folder in Outlook makes the server build a list of every mail in it before it can answer. On a 400,000-mail folder that took about 6.8 seconds for the normal date-sorted view and about 34 seconds with conversation view on, and a view filtered on unread took about 10 seconds. Outlook shows "not responding" for that whole time.

Ten changes. Each was measured on its own, and each was checked to leave the resulting table exactly as it was.

## The changes

1. **Do not build the whole list just to count it.** To say how many mails a folder holds, the server lined every mail up, counted the line and threw it away; counting the mails directly gives the same number. Search folders genuinely need the line-up and keep it. `ropGetContentsTable` 5292 ms to 80 ms.
2. **Answer "where am I in the list?" without building the list.** The cursor position is already tracked and the folder size is one cheap count. `ropQueryPosition` 2517 ms to 41 ms.
3. **Switch on the statement cache while loading.** The load asks for one property at a time, hundreds of thousands of times, and each ask was explained to the database from scratch. `query_content` already takes this cache for the same reason.
4. **Ask "is this mail read?" with a placeholder** instead of writing the mail's number into the query text, which stopped the database ever reusing it. 4.5 us down to 2.5 us, once per mail.
5. **Reuse the per-group queries.** Grouping mails asks almost the same question once per group, 400k times; it only varies with grouping depth. 2176 ms to 23 ms.
6. **Build the ephemeral table's indexes after the rows are in,** rather than filing every row into six of them on the way. Inserting rows 6101 ms to 2454 ms, plus 1205 ms to build the indexes.
7. **Take group unread counts from the grouping query.** Each header was written, then all its mails walked to count unread ones, then the header corrected. The grouping query can sum them itself. 19.76 s to 18.66 s over six runs.
8. **Insert rows in batches of 128** where the numbering is predictable. Restricted, filtered and sorted loads keep the row-at-a-time path, because they can skip rows and so cannot predict it.
9. **Cache the attachment and association probes.** Both wrote the mail's number into the query text, so a filtered load paid 800k query preparations. 10.50 s to 7.79 s.
10. **Only work out the message flags a filter actually looks at.** PR_MESSAGE_FLAGS is not stored as one value; it costs one lookup per flag bit. A filter on unread reads a single bit but paid for all six on every mail. 7.42 s to 1.75 s.

## Result

Server time for one folder open, `rop_debug = 2`, Outlook client, 399,986-mail folder:

| view | before | after |
|---|---|---|
| normal, date sorted | 6794 ms | about 1770 ms |
| conversation view | 33633 ms | about 20500 ms |
| filtered on unread | 10504 ms | 1750 ms |

What remains in conversation view is one build of a 400,000-group table, which is what that view asks for rather than a missing optimisation. The test mailbox holds one mail per conversation, which is the worst case for it; real mailboxes have fewer, larger conversations.

## Checking nothing changed

Every change was checked by hashing each column of each row of the resulting table and comparing with the commit before it, over six shapes: 501 groups of about 800 mails, 3 groups of about 133k, 400k single-mail groups, each collapsed and expanded, plus the flat and the filtered tables. Hashes are compared in row order and order-independently, and the columns include the computed ones (PR_MESSAGE_FLAGS, PR_HASATTACH, PR_ASSOCIATED, PR_CONTENT_UNREAD) so that a change in how those are worked out would show up. All identical.

One gap worth stating: the public-store adjustment that forces the read flag moves into cu_get_msg_flags in the last commit. The test store is a private one, so that line is verified as code motion rather than by measurement.

### DIFF
--- a/exch/exmdb/table.cpp
+++ b/exch/exmdb/table.cpp
@@ -927,12 +927,77 @@ static bool table_load_content_table(db_conn &db, db_base_wr_ptr &dbase,
 	auto cl_optim = HX::make_scope_exit([&]() { db.end_optim(); });
 
 	/*
+	 * An unsorted table without a restriction takes every row the scan
+	 * yields, in scan order, so rows can go in several at a time. row_id is
+	 * assigned consecutively by AUTOINCREMENT, which prev_id and idx are
+	 * derived from, so both are known before the insert.
+	 */
+	bool rows_done = psorts == nullptr && conv_id == nullptr &&
+	                 prestriction == nullptr;
+	if (rows_done) {
+		auto stm = db.eph_prep(fmt::format("SELECT IFNULL(MAX(row_id), 0) FROM t{}", table_id));
+		if (stm == nullptr || stm.step() != SQLITE_ROW)
+			return false;
+		uint64_t next_row = stm.col_uint64(0) + 1;
+		stm.finalize();
+
+		static constexpr unsigned int batch = 128;
+		auto row_tpl = fmt::format("(?, ?, {}, 0, 0, ?)", CONTENT_ROW_MESSAGE);
+		auto head = fmt::format("INSERT INTO t{} (inst_id, prev_id, "
+		            "row_type, depth, inst_num, idx) VALUES ", table_id);
+		std::string batch_sql = head;
+		for (unsigned int i = 0; i < batch; ++i)
+			batch_sql += (i == 0 ? "" : ", ") + row_tpl;
+		auto bstm = db.eph_prep(batch_sql);
+		if (bstm == nullptr)
+			return false;
+
+		std::vector<uint64_t> mids;
+		mids.reserve(batch);
+		auto flush = [&]() {
+			sqlite3_stmt *st = bstm;
+			xstmt tail;
+			if (mids.size() != batch) {
+				std::string s = head;
+				for (size_t i = 0; i < mids.size(); ++i)
+					s += (i == 0 ? "" : ", ") + row_tpl;
+				tail = db.eph_prep(s);
+				if (tail == nullptr)
+					return false;
+				st = tail;
+			} else {
+				sqlite3_reset(st);
+			}
+			for (size_t i = 0; i < mids.size(); ++i) {
+				auto row = next_row + i;
+				sqlite3_bind_int64(st, i * 3 + 1, mids[i]);
+				sqlite3_bind_int64(st, i * 3 + 2, row - 1);
+				sqlite3_bind_int64(st, i * 3 + 3, row);
+			}
+			if (gx_sql_step(st) != SQLITE_DONE)
+				return false;
+			next_row += mids.size();
+			mids.clear();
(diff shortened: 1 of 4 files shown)


exch/exmdb/table.cpp
```diff
@@ -326,22 +328,56 @@ static std::string table_cond_to_where(const std::vector<condition_node> &list)
...
+using tlc_stmt_cache = std::map<uint64_t, xstmt>;
+
+static uint64_t tlc_stmt_key(const std::vector<condition_node> &cond_list,
+    int depth, bool leaf)
+{
+	uint64_t nullmask = 0;
+	size_t i = 0;
+	for (const auto &cnode : cond_list) {
+		if (cnode.pvalue == nullptr)
+			nullmask |= static_cast<uint64_t>(1) << i;
```
jengelh, line 352: I have cherry-picked the first four commits.

1<<i only leaves room for 64 conditions. A switch to e.g. `HXbitmap_set` would be needed.

jengelh (changes requested):

r4nc0r:
Thanks for picking up the first four.

On the 64-condition limit: you are right, and there turned out to be a worse one behind it. `psort[]` holds `count` entries, `rop_sorttable` caps `count` at `MAXIMUM_SORT_COUNT` and `table_load_content_table` caps it at 16 -- but nothing validated `ccategories` against `count`, and `table_load_content` indexes `psort[depth]` for every `depth < ccategories`. An oversized `ccategories` was therefore already reading past `psort[]`, before any shift could overflow. The mask made a pre-existing out-of-bounds read louder rather than introducing the only problem there.

So the first commit now rejects `ccategories > count` at the exmdb entry, gives the array bound a name (`gct_max_sorts`), and static_asserts that the mask is wide enough for it. The recursion depth is then provably at most 16 rather than merely usually so, which is what makes the `uint64_t` sufficient for everything a caller can ask for.

If you would rather the mask not depend on that bound at all, say so and I will switch it to `HXbitmap_set`. I went this way because the guard fixes the `psort[]` read as well, and once the bound is actually enforced a heap bitmap for at most 16 bits looked like the wrong trade.

The branch is rebased on prime with those four dropped, so it is six commits now. Two of them touched functions you renamed, so they are rewritten rather than rebased: they use `cu_msg_is_fai` and `cu_msg_has_attachments`, and the categorized-table commit picks up `tablesort_is_minmax`. Your `bool` over `BOOL` corrections are applied to the rest of the series too.

The attachment/association caching is meant to sit on top of your `SELECT 1 ... LIMIT 1` change rather than replace it: that one stops the query walking every attachment row, this one stops it being prepared afresh for every message.

One gap I should flag rather than let you find: the public-store adjustment that forces the read flag moves into `cu_get_msg_flags` in the last commit. Both stores I test against are private, so that line is verified as code motion and never actually executed.

PR #316, opened 2026-08-13
ews: enforce Send-As permission on client-supplied From/Sender

nourdineb-ops:
Sorry, missed something in #286 that I should have caught then. Found and fixed a permission gap in EWS's `CreateItem` while looking into something unrelated (a "Send As" convenience script on our own deployment). `toContent()` writes whatever identity a client supplies via a message's From/Sender field straight into `PR_SENT_REPRESENTING_*`/`PR_SENDER_*`, for both `SaveOnly` and `SendOnly`/`SendAndSaveCopy` - with no authorization check.

Root cause: #286 ("respect client-supplied From on save") intentionally started trusting a client-supplied From to fix a real bug - a draft saved from a secondary alias reverting to the primary identity - and explicitly modeled this on grommunio-web's `PR_SENT_REPRESENTING_* ??= PR_SENDER_*` pattern (`class.operations.php`). It only carried over the fallback-default half of that pattern, not the permission check that makes it safe there.

Fix adds `EWSContext::checkSendAs()`, mirroring `zs_submitmessage()`'s check (`frightsGromoxSendAs` or full store ownership on the claimed identity's mailbox), called from `CreateItem` right after `toContent()` populates `PR_SENT_REPRESENTING_EMAIL_ADDRESS` - applies uniformly to drafts and actual sends, and to the reply/forward wrapper types alongside plain messages.

Tested against a live deployment: granted a real `frightsGromoxSendAs` grant between two test mailboxes, confirmed a `CreateItem` with the granted identity still succeeds and correctly resolves to it, confirmed the same request with an ungranted identity now gets `ErrorAccessDenied`, and confirmed a plain message with no From override is unaffected.

### DIFF
--- a/exch/ews/context.cpp
+++ b/exch/ews/context.cpp
@@ -3314,6 +3314,35 @@ uint32_t EWSContext::permissions(const std::string& maildir, uint64_t folderId) 
 }
 
 /**
+ * @brief      Verify the caller may claim a given identity as sender
+ *
+ * @param      identity   SMTP address the message claims to be sent from
+ *
+ * @throw      EWSError::AccessDenied   Caller lacks Send-As rights on
+ *              the claimed identity
+ */
+void EWSContext::validate_sendas_perms(const std::string &identity) const
+{
+	if (identity.empty() ||
+	    strcasecmp(identity.c_str(), m_auth_info.username) == 0)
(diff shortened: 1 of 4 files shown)


nourdineb-ops:
Just checking in - any questions on this one, or anything else you'd like me to test/expand on before it gets a look?

PR #318, opened 2026-08-14
oxcmail: do not drop HTML bodies that are not the first subpart

r4nc0r:
`select_parts()` only adopts HTML parts when the **first** child of a non-`multipart/alternative` container has already yielded one:

```c
if (child_idx == 0 && cld_info.htmls.size() > 0)
        hjoin_enabled = true;
...
if (hjoin_enabled) {
        info.htmls.insert(...);
        info.hjoin.insert(...);
}
```

`hjoin_enabled` is meant to select whether *multiple* HTML parts get joined (introduced in aa745b88e for the EXC-style multi-HTML case), but it also gates whether *any* HTML is picked up at all. An HTML body preceded by anything else — an attachment, a `text/plain` part, an inline image — is discarded and demoted to an attachment. `PR_HTML` is then autosynthesized from the plaintext body and wrapped in `<pre>`, so the message presents as plaintext. Where no `text/plain` part exists either, the message ends up with **no body at all**.

Containers whose first child carries the HTML are unaffected, which is why this is not more widely visible.

## Affected shapes

Measured against `tests/oxcmail_ie` before the change:

| MIME structure | before | after |
|---|---|---|
| `mixed{ alternative{plain,html}, pdf }` | HTML ok | HTML ok (unchanged) |
| `mixed{ pdf, alternative{plain,html} }` | `<pre>` plaintext, HTML → attachment | HTML ok |
| `mixed{ plain-disclaimer, alternative{plain,html} }` | `<pre>` plaintext, HTML → attachment | HTML ok |
| `related{ inline image, alternative{plain,html} }` | `<pre>` plaintext, HTML → attachment | HTML ok |
| `mixed{ plain, html }` | `<pre>` plaintext, HTML → attachment | HTML ok |
| `mixed{ pdf, html }` | **no body at all** | HTML ok |

### DIFF
--- a/lib/mapi/oxcmail2.cpp
+++ b/lib/mapi/oxcmail2.cpp
@@ -117,6 +117,16 @@ void select_parts(const MIME *part, MIME_ENUM_PARAM &info, unsigned int level) t
 		if (hjoin_enabled) {
 			info.htmls.insert(info.htmls.end(), std::make_move_iterator(cld_info.htmls.begin()), std::make_move_iterator(cld_info.htmls.end()));
 			info.hjoin.insert(info.hjoin.end(), std::make_move_iterator(cld_info.hjoin.begin()), std::make_move_iterator(cld_info.hjoin.end()));
+		} else if (info.htmls.empty() && !cld_info.htmls.empty()) {
+			/*
+			 * The HTML body is not required to sit in the very
+			 * first subpart; an attachment, a text part or an
+			 * image can precede it. Joining is not applicable
+			 * here (that is what hjoin_enabled selects), but the
+			 * body must still be picked up rather than dropped.
+			 */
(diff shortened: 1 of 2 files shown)


jengelh:
please have a look at refs/heads/mailat meanwhile (almost no description, but test_6 will work)

jengelh:
added 797e8301e7ce188bcd59d35ca5ccd7878f431cb6 and 4cfae31b8455027e20871d47e7e32f8153b4e4d2.

Your testcase T8 is an IMO invalid interpretation of MIME. {PR_BODY, PR_HTML, PR_RTF_COMPRESSED} is supposed to be a set of alternatives. This requires multipart/alternative{text, html}. multipart/mixed{text, html} is something different.

PR #322, opened 2026-08-18
mapi: missing nullptr check added

EugenePopovych:
Missing check for nullptr result of event->get_line() call.
Fixing coverity issue 1663925.


EugenePopovych:
I assume it may have sense to add some error code into errstr but I'm not sure what this code should be and if I have to register or mention it anywhere.

jengelh:
No errstr needed. A file could be specially crafted to just lack the line.

PR #334, opened 2026-08-25
exch: small thread-safety improvement in emsmdb

EugenePopovych:
Move-constructor for notification_ctx is made thread-safe.
Fixing coverity issues 1685359, 1685358, 1685357, 1663976, 1663972.

### DIFF
--- a/exch/mh/emsmdb.cpp
+++ b/exch/mh/emsmdb.cpp
@@ -83,7 +83,7 @@ struct EMSMDB_HANDLE2 : public EMSMDB_HANDLE
 
 struct notification_ctx {
 	notification_ctx() = default;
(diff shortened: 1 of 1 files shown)


jengelh:
The locking is technically permitted to throw, so `noexcept` is no longer valid.

jengelh:
Locking a mutex inside a move constructor feels wrong.
Instead of fighting the language to make a thread-safe object movable, there are some options:
1. separating thread-safety from data
2. pointer indirection
3. deleting move/copy

I went with 2/3 in 4c1cb9bd0fd86481bcca0b6d0740eb0211484a79, short of figuring out how to do 1 (if that's even worth spending time on).

PR #337, opened 2026-08-31
lda: separate the autoreply subject prefix from the subject

r4nc0r:
autoreply_subject_prefix is concatenated with the subject of the incoming message, so the prefix has to carry the separating space itself. The compiled-in default does, but config_file_init trims trailing whitespace off every configuration line, so an administrator cannot write one. A prefix set in gromox.cfg therefore produced

	Subject: Out of Office:Re: Vertrag

Append a space to a configured prefix that does not end in whitespace. The default is unaffected, and the branch that drops the separator when there is no incoming subject still finds it.

The default itself said "Out of office" where the manual page, the rest of the documentation and grommunio-web all say "Out of Office". Spell it the same way, so a mailbox left on the default and one whose subject grommunio-web filled in agree.

### DIFF
--- a/doc/gromox.cfg.5
+++ b/doc/gromox.cfg.5
@@ -27,7 +27,8 @@ Out-of-office replies use the subject that was set alongside the OOF message.
 Clients which configure OOF over EWS cannot set one, because
 SetUserOofSettings has no field for it. When no subject is stored, this prefix
 plus the subject of the incoming message is used instead. Set to the empty
-string to send such replies without a subject.
+string to send such replies without a subject. A prefix not ending in
+whitespace is separated from that subject by a space.
 .br
 Default: \fIOut of Office: \fP
 .TP
--- a/mda/exmdb_local/exmdb_local.cpp
+++ b/mda/exmdb_local/exmdb_local.cpp
@@ -14,6 +14,7 @@
 #include <unistd.h>
 #include <vector>
 #include <fmt/core.h>
+#include <libHX/ctype_helper.h>
 #include <libHX/string.h>
 #include <sys/stat.h>
 #include <gromox/bounce_gen.hpp>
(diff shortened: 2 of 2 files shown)


ITaluone:
Can I chip in with an additional fix request here?

MS Exchange does also include the original mail subject in the OOF reply mail.

orig subject: "Things regarding vacation"
grommunio's (gromox 3.9.17.a5462c9-lp156.1.1) reply mail's subject: "Out of office" 
Exchange's OOF mail subject: "Automatic reply: Things regarding vacation" (at least I recall seeing this for Exchange 2016 back then)

r4nc0r:
That is already in, and it is what this PR is a follow-up to.

[#326](https://github.com/grommunio/gromox/pull/326) added `gromox.cfg:autoreply_subject_prefix`. When a mailbox has no out-of-office subject stored, the reply subject becomes that prefix followed by the subject of the incoming message, so your example comes out as `Out of office: Things regarding vacation`.

The wording is yours to choose, since the prefix is a plain string. If your Exchange used `Automatic reply: `, then

```
autoreply_subject_prefix=Automatic reply: 
```

gives you exactly that. The compiled-in default is `Out of office: ` on `master` today, and this PR spells it `Out of Office: ` to match the manual page and the rest of the documentation. Setting it to the empty string turns the subject off again.

One thing that will otherwise look like the feature not working

The derivation only happens when **no** subject is stored on the mailbox. A stored one always wins.

That matters depending on how the out-of-office was switched on. Configured over EWS, which is what Outlook does, no subject is stored, because `SetUserOofSettings` has no field for one, and the derivation applies. Configured in grommunio Web, a subject *is* stored, and the derivation is skipped for that mailbox no matter what the prefix is set to.

So if you try this and still see a bare subject, check whether the mailbox has one stored from a previous grommunio Web visit. There is a web-side change for that at [grommunio-web#139](https://github.com/grommunio/grommunio-web/pull/139), which stops it writing a fixed subject and shows the server's prefix in the form instead.

PR #339, opened 2026-09-03
Per-domain SMTP gateway table (smart-host schema)

gott-alexander:
## Summary

Adds the `domain_smtp_gateway` table to the dbop MySQL schema (version 134)
and documents how a sender-dependent MTA evaluates it. The table lets each
hosted domain define its own outbound relay (smart-host) — gromox owns the
database schema, so the table is created and migrated here.

## Changes

- `lib/dbop_mysql.cpp`: add `tbl_dsgw_134` and register it in
  `tbl_init_top` and `tbl_upgrade_list` as schema version 134. Fresh
  installations (`gromox-dbop -C`) create the table with all other tables;
  existing installations are migrated transparently (`gromox-dbop -U`, or
  automatically via `mysql_adaptor schema_upgrade`). One row per domain,
  foreign key to `domains.id` with cascade delete, columns for host, port,
  encryption mode, optional SASL credentials, enabled flag and a
  description. The encryption column is a plain varchar (validated by the
  management layer) rather than an ENUM, so new modes do not require a
  schema migration.
- `doc/mta-smart-hosts.rst`: document the table layout and the intended
  evaluation. Outbound relay selection deliberately lives in the MTA, not
  in gromox: gromox hands mail to its `outgoing_smtp_url` (default
  `sendmail://localhost`) and the MTA resolves the relay per envelope
  sender domain. This keeps queueing, retry, TLS policy and DSN handling
  in one place instead of duplicating an SMTP client stack inside gromox.
  The document includes a complete Postfix wiring example
  (sender-dependent transport map, sender-dependent SASL map, TLS
  transport services) and the caveats (Postfix `encrypt` level does not
  verify certificates; implicit TLS expects a wrappermode port).

## Verification

- `gromox-dbop -C` on an empty database creates all tables including
  `domain_smtp_gateway`, `schemaversion` = 134.
- `gromox-dbop -U` on a schema-133 database migrates to 134 and creates
  the table; downgrade path unaffected.
- The Postfix wiring from the documentation was tested end to end
  (authenticated delivery via STARTTLS relay, all encryption variants,
  fallback for domains without a gateway row).


gott-alexander:
I'll create another update that uses dbop_mysql.cpp instead of the SQL file.

PR #340, opened 2026-09-08
http: fix shutdown-order race causing segfaults in hpm_processor_insert_ctx

nourdineb-ops:
`gromox-http` segfaulted repeatedly during shutdown/restart, always backtracing to `hpm_processor_insert_ctx()` indexing `g_context_list` on already-invalid state, called from `http_context::~http_context()` while the `http_context[]` array was being torn down.

Root cause: the connection-accept thread (`listener_trigger_accept()`, via `listener_ctx::watch_start()`) was stopped **last** during shutdown. Its `HX::make_scope_exit(listener_stop)` guard was registered right after `listener_init()`, near the top of `main()` — so in the LIFO unwind it fired dead last, after `threads_pool_stop`, `contexts_pool_stop`, `http_parser_stop` (destroys the `http_context[]` array, calling `hpm_processor_insert_ctx` per context) and `hpm_processor_stop` (resets its own context array to null).

That left the accept thread free to keep running throughout the teardown of everything it hands connections to — a connection accepted (or already queued) in that window could be inserted into `contexts_pool` and processed into context arrays that were being destroyed or had already been reset, racing the teardown.

Fix: keep `listener_init()` (socket bind) where it was, but move the `scope_exit` registration for `listener_stop` to after `threads_pool`'s, so it is the *first* thing torn down on shutdown — no more connections are accepted or handed off before `contexts_pool`/`http_parser`/`hpm_processor` are touched at all.

Built and deployed to production; a full service restart cycle across all 9 gromox daemons on that host came up clean afterward with no recurrence.

### DIFF
--- a/exch/http/main.cpp
+++ b/exch/http/main.cpp
@@ -292,7 +292,11 @@ int main(int argc, char **argv)
 	mlog(LV_INFO, "http: fastcgi execution timeout is %s", temp_buff);
 	if (listener_init(*gxconfig, *g_config_file, http_support_tls) != 0)
 		return EXIT_FAILURE;
-	auto cleanup_4 = HX::make_scope_exit(listener_stop);
+	/*
+	 * cleanup_4's scope_exit is deliberately NOT registered here even
+	 * though listener_init() just ran - see below, right after
+	 * threads_pool's cleanup_28, for why.
+	 */
 
 	const char *program_identifier = "http";
 	auto istore_standalone = gxconfig->get_ll("istore_standalone");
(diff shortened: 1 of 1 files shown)


nourdineb-ops:
Note on the failing CI check: unrelated to this change — the repo's `build` workflow has been failing at the `install-deps`/GnuTLS step (cascading into vmime's CMake configure) since at least the 2026-09-03 push to `master` itself, not just PRs. Happy to rebase/retrigger once that's sorted.

nourdineb-ops:
Following up — a week with no review. The failing CI check is unrelated to this change (see comment above — the `install-deps`/GnuTLS step has been broken repo-wide since before this PR).

PR #342, opened 2026-09-09
ews: flush the parent message instance after CreateAttachment

nourdineb-ops:
Attachments added via `CreateAttachment` were silently lost by the time the message was sent: reproduced live (Outlook for Mac, compose -> attach a file -> send), confirmed both in the delivered mail and in the Sent Items copy - and directly in `exchange.sqlite3`, where the freshly sent message had zero rows in `attachments` despite `CreateAttachment` having reported `ResponseClass="Success"` moments earlier.

Root cause: `CreateAttachment`'s handler flushes only the newly-created *attachment* instance (`flush_instance(aInstId)`), never the *parent message* instance (`mInst->instanceId`) it belongs to. Per `exmdb_server::flush_instance()`'s `instance_type::attachment` branch (`exch/exmdb/instance.cpp`), flushing an attachment instance only merges it into its parent message instance's in-memory `MESSAGE_CONTENT` - it never touches the database. The attachment only becomes durable once the *message* instance itself is flushed, and nothing in `CreateAttachment`'s handler ever did that. `DeleteAttachment`, right below it in the same file, gets this right: it flushes `mInst->instanceId` after modifying the message instance's attachments.

Fix: flush the parent message instance too, mirroring `DeleteAttachment`.

Verified live (Outlook for Mac): attachment now present in `attachments`/`attachment_properties` for the sent message, and the delivered mail (checked via the SMTP relay log) carries the expected size increase for the attached file.

### DIFF
--- a/exch/ews/exceptions.hpp
+++ b/exch/ews/exceptions.hpp
@@ -603,6 +603,7 @@ E(3471, "failed to delete cancelled meeting");
 E(3472, "cannot read from item's source folder");
 E(3473, "cannot write to destination folder");
 E(3474, "access denied: missing Send-As permission on the claimed sender identity");
+E(3475, "failed to persist attachment to parent message");
 
 #undef E
 }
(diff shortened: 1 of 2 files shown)


nourdineb-ops:
Just flagging this one's open — small single-commit fix (flush the parent message instance after CreateAttachment so subsequent reads see the new attachment).

Issue #350, opened 2026-09-11
No notifications for empty folder


jengelh:
Version: gromox-3.10-93-g1f529d2da , gromox.cfg:istore_standalone=3 (!)

Using MFCMAPI, open the contents table of a folder that is empty (e.g. PRIVATE_FID_ROOT in a blank Gromox mailbox), then execute action (menubar) > Folder > New > Generic message. The newly created message is not shown in the contents table. My debugging pinpoints to gromox-3.9-247-g 7711caccdaf3b91cab7f3400fdf196514c869bd9 (@r4nc0r) which changed:

```diff
+       if (pfolder->type != FOLDER_SEARCH)
+               return ecSuccess;
        auto err = rtable->load();
```

`rtable->load()` is no longer called, thus no `exmdb_callid::load_content_table` EXRPC is made, thus no MAPI table object (in exmdb's process image / tables.sqlite3) is created, thus `dbase.tables.table_list` is empty. When `emxdb_server::write_message` executes as part of "New generic message", control eventually lands in `dbeng_notify_cttbl_add_row`. This function observes a `dbase.tables.tables_list` with 0 elements and the notification goes nowhere.

```c++
        auto sql_transact_eph = gx_sql_begin(pdb->m_sqlite_eph, txn_mode::write);
        if (!sql_transact_eph) {
                mlog(LV_ERR, "E-2063: failed to start transaction in cttbl_add_row");
                return;
        }
        for (auto &tnode : dbase.tables.table_list) { // <-- here
                auto ptable = &tnode;
```

The bandaid seems to entail reverting 7711caccdaf3b91cab7f3400fdf196514c869bd9 .

r4nc0r:
Confirmed, and 7711caccd is half of it. The other half is 69657ddf9, which lets QueryPosition answer from a count without loading either, so a client that reads the row count and sees 0 never issues a ROP that would create the exmdb table. Both rely on the same assumption: a client with rows to show has to fetch them, and the fetch loads the table before it can observe anything. An empty table is the one case where nothing forces a load, and the only way the client can learn of the first row is the notification that needs the table to exist.

[...]

PR #351, opened 2026-09-11
emsmdb: keep an empty content table loaded so notifications reach it

r4nc0r:
## What happens

Open the contents table of an empty folder with MFCMAPI and create a message in it. The message does not appear in the table. Reported in #350 against gromox-3.10-93-g1f529d2da.

## Why

exmdb addresses its table notifications to the table objects it holds; `dbeng_notify_cttbl_add_row` walks `dbase.tables.table_list` to find them. A content table that emsmdb has not loaded has no such object, so a message created in that folder produces no notification for the client's table handle.

Two commits made that load lazy: 7711caccd stopped `ropGetContentsTable` from building the table to count it, and 69657ddf9 lets `ropQueryPosition` answer from a count without loading. Both rely on the client fetching rows, which loads the table. A client with rows to show has to do that before it can observe anything. An empty table is the case where nothing forces a load: there are no rows to fetch, and the only way the client can learn of the first one is the notification that needs the table to exist.

## The change

An empty content table is loaded right away, on open, when a sort unloads it, and when its position is queried. `table_object::load_if_empty` does the second and third; `ropGetContentsTable` uses the count it already has and does so ahead of the `MAPI_DEFERRED_ERRORS` early return, so a client that opens with that flag gets the table too, which that path never did. Loading an empty table costs one indexed count and an empty scan. Populated tables keep the deferred load and its savings, and search folders keep loading on open as before.

A table that a restriction leaves empty is still loaded only when the client asks for rows or position, because knowing that it is empty would require the load in the first place.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Enhancements:
 
 Fixes:
 
+* emsmdb: Opening a folder that has no contents no longer misses notifications.
 * The rule processor executes Extended Rules' OP_MOVE/OP_COPY now and matches
   string restrictions across PT_STRING8↔PT_UNICODE.
 * During MAPI-to-iCalender conversions, TZID is now omitted when exporting
--- a/exch/emsmdb/oxcfold.cpp
+++ b/exch/emsmdb/oxcfold.cpp
@@ -875,6 +875,14 @@ ec_error_t rop_getcontentstable(uint8_t table_flags, uint32_t *prow_count,
 		return aoh_to_error(hnd);
 	rtable->set_handle(hnd);
 	*phout = hnd;
+	/*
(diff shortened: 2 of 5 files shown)


doc/changelog.rst
```diff
@@ -12,6 +12,8 @@ Enhancements:
 
 Fixes:
 
+* emsmdb: Opening a folder (content table) that has no contents missed
+  getting notifications, which was fixed.
```
crpb, line 16: ", which was fixed" can be omitted in the "Fixes:" section

jengelh: Generally, I like having complete sentences (i.e. with a verb). I'll reword it before merge.

